PR #230, opened 2022-10-06
Add databse schemas

bendnorman:
I created database schemas for the data warehouse and data mart tables using SQL Alchemy models. This removes the need for pandera.

I only ran into a couple of database constraint errors. Questions for @TrentonBush:
1. `iso_projects_2021`: What is the difference between `queue_id` and `project_id`? why are there multiple records for a single `queue_id`? 
2. `iso_locations_2021`: What is the pk here? Can a project have multiple locations? Some projects have two locations. some of these have the same `county_id_fips`, but different `raw_county_names`. Is this because of the county_1, county_2 … thing? What does it mean for a project to be in two counties? 
3.  Are we still using `iso_for_tableau`? 
4. `eip_projects`, `eip_facilities`: What is `ccs_id`? Does it require a constraint or a separate table? 
5.  `eip_facility_project_association`: So is there a many-to-many relationship between projects and facilities? How are facilities different from projects? Currently there are `project_ids` in `eip_facility_project_association` that don’t exist in `eip_projects`.
6. `eip_project_permit_association `: Both foreign key constraints failed. 
7. Do `state_policy` and `local_ordinance` have a natural key? I guess not because a state or county could have multiple ordinances. 
8. Is there a natural key for `contested_project`? Maybe `project_name`, but it is missing a bunch of values.

Things that need to be resolved before merging this in:
- [ ] `eip_project_permit_association` fk errors
- [ ] `eip_facility_project_association` fk errors


TrentonBush:
1. `iso_projects_2021.queue_id` comes from the raw data and I could never figure out how to make a useful ID field from it. There are duplicates because I think each source ISO has their own. But even a compound key of (ISO, queue_id) was not unique. Maybe it would be clearer to rename it `raw_queue_id` or something to indicate it is not ready for direct use?
2. The primary key of `iso_locations_2021` *should* be (`project_id`, `county_id_fips`). Some projects have multiple counties, but I don't actually know what that means. It is only true for proposed projects (maybe they don't know the exact location yet?). Yes, it comes from the raw data having "county_1, county_2, etc". Right now it looks like that PK doesn't quite work because project 7295 has a null raw_state_name leading to null county_id_fips, and project 5500 has two different `raw_county_name` entries that actually map to the same county (one is a county, the other is a town within that county). I guess we'll have to fix those before assigning that PK.
3. No, `iso_for_tableau` is not used and should be removed whenever we do a repo cleanup.
4. `ccs_id` is new as of the latest data update. It does map to a separate table in the raw data, but that table is not part of the current ETL process. Maybe we should rename it `raw_ccs_id`
5. Non-matching IDs (I see 8 `project_id`s and 5 `facility_id`s) are a quality issue in the raw data we'll have to deal with. The difference between a facility and a project is that a project is a piece of construction while a facility is a discrete industrial plant. The relationship is many-to-many because one project (like a new pipeline) may touch multiple facilities, and one facility can have multiple proposed construction projects. But most of them are one-to-one.
6. Similar to above, but with a lot more missing data (186 project_ids, 25 air_construction_ids)
7. I guess the natural key for `local_ordinance` is actually (`geocoded_locality_name`, `raw_state_name`). I see two whole-row duplicates (Brownsville and Benbrook, TX) that currently prevent this being a PK. The jurisdictional granularity of this data is a mix of counties and cities/towns. So `county_id_fips` in this table should actually be interpreted as "county_id_fips that contains this jurisdiction".
8. We don't use this table currently, but I think it needs a surrogate key if there are a bunch of null `project_name`s

bendnorman:
1. I think it makes sense to rename it to `queue_id_raw`. Will this break any dashboards? Also, what is the impetus behind keeping around `_raw` columns? 
2. It looks like there are a handful of geocoding failures that make it difficult to enforce this PK right now. Also, instead of having an `iso_locations_2021` table, why not just add state_id_fips and county_id_fips codes to the `iso_projects_2021` table? 
3. Ok, I won't create a schema and will remove the table in a future PR.
4. Renaming sounds good to me. Do you expect CSS data to be important in the future? 
5 - 8. Should I just create issues for these data cleaning tasks so we can merge this PR in?

TrentonBush:
1. No, the dashboards only touch tables in the `data_mart` schema, and no `data_mart` tables use `queue_id`. Also, the dashboards run on cached data (currently set up to be manually refreshed) so even a bad push to BigQuery won't break them automatically.
    1. I kept the `_raw` columns 1) for convenience when debugging derived columns 2) for manual use when we want to see more detail on a particular row without committing to a big QC process for the entire column (the street address for a particular plant, for example, is human readable but not machine standardized, and we don't care to look at it often enough to justify cleaning all of them)
2. Looking at it again, some projects aren't even in the US, so `county_id_fips` will never work as part of the PK. Maybe we have to use the `raw_county_name`? This table is separate from `iso_projects_2021` because of the one-to-many relationship between projects and counties. But there are only 36 multi-county projects (out of 7642).
3. Sounds good
4. I don't expect CCS data to ever be important
5+. Sure. Does this PR attempt to enforce these schemas? That would halt the load process due to all these bad PK/FKs, right?

bendnorman:
## Removing Pandera
@TrentonBush I added SQLAlchemy models for the data mart tables. There are clear primary keys for all of the tables, but some of the columns in the natural keys are missing values. I've created an issue enumerating the issues. 

I also removed the need for pandera validation by using pandas `convert_dtypes()` function. This means we can delete the `schemas` module and remove pandera as a dependency. I wanted to check in with you before I do this because there are a few notebooks that depend on the pandera schemas.

- 02.1-kl-deduplicate-active-iso-queue.ipynb
- 02-kl-deduplicate_active_iso_queue.ipynb
- parish.ipynb
- mcoe.ipynb

## Testing out new schemas
This PR shouldn't have changed any of the actual data, but some types might have changed. I created a new GCP project called `deployment-gap-model-dev` so I could test out the new schemas in BigQuery without clobbering our "production" data. @TrentonBush do you think there is an easy way to test out the dashboard with the dev data before we update the production data?

## TODO before merging this PR in
- [ ] Remove schema module and pandera library
- [ ] Remove unused metadata directory and rename `models` module to `metadata` for naming consistency.
- [ ] Create an issue for data quality issues discovered in the data warehouse and mart.
- [ ] Create an issue for moving data dictionary information into the SQL alchemy models.
- [ ] Test the dashboards on the data from this PR.

TrentonBush (approved):

Issue #250, opened 2022-12-13
Remove obsolete tables


TrentonBush:
The following tables are no longer used
### Data Warehouse
- [x] ISO 2020 data
    - [x] `iso_locations`
    - [x] `iso_projects`
    - [x] `iso_resource_capacity`
    - [x] `iso_for_tableau` (no longer generated but still lingering in BigQuery)

Note that there is some coupling/duplication of code from 2020 and 2021 processing code. See issue #206 for example.
### Data Mart
- [x] `co2_dashboard`
- [x] `proposed_power_dash_existing_plants`
- [x] `proposed_power_dash_proposed_plants`

Note that `proposed_power_dash_local_opp` **is still used**. Ideally it would be refactored to a SQL query but that should be a separate issue.

PR #251, opened 2022-12-14
Remove unused code

bendnorman:


bendnorman:
I removed everything in #250 except for the co2_estimates table because the `counties` data mart table depends on code in the `co2_dashboard` module. I could delete the `co2_estimates` module by moving the private functions `counties` uses to the helpers module or the `counties` module

TrentonBush:
I made one small change to `src/dbcp/data_mart/__init__.py` to stop it from creating the defunct `co2_dashboard` table

TrentonBush (approved):

PR #282, opened 2023-06-01
Create a dev BigQuery dataset

bendnorman:
This PR creates two new BQ datasets in GCP project `dbcp-dev-350818`: `data_warehouse_dev` and `data_mart_dev`. These new development datasets will allow us to push changes to BQ without breaking downstream services (google sheets and tableau).

Things I'd like to do when I get back from vacation.
- [ ] Rename our "production" datasets from `data_warehouse` and `data_mart` to `data_warehouse_prod` and `data_mart_prod`. @TrentonBush this will likely require a decent amount of Tableau changes. (We've decided to pass on this for now.)
- [x] Adjust the IAM so the dev datasets are hidden from folks in the `big-query-users` group. I'd like to do this once the `prod` datasets have been renamed and I can test the permissions with a climate partners person.
- [x] Delete the old `deployment-gap-model-dev` gcp project.

I figured this PR can be merged in now and the above issues could be handled in a separate PR.

### DIFF
--- a/makefile
+++ b/makefile
@@ -7,20 +7,29 @@ shell:
 etl_local:
 	docker compose run --rm app python -m dbcp.cli --etl
 
-etl_bq:
-	docker compose run --rm app python -m dbcp.cli --etl --upload-to-bigquery
-
 data_mart_local:
 	docker compose run --rm app python -m dbcp.cli --data-mart
-
-data_mart_bq:
-	docker compose run --rm app python -m dbcp.cli --data-mart --upload-to-bigquery
 
 all_local:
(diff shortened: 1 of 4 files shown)


TrentonBush:
Nice! Thanks for setting this up. I tested out the changes and they look good to me.

All the post-vacation to-dos sound good too. Is the point of the proposed `_prod` suffix basically just to make dev/prod more explicit? I guess we should probably do it but I'm looking for an excuse to be lazy.

TrentonBush (approved):

bendnorman:
Yeah, the name change would just make things more explicit. We should probably scope out how much work it would be to update all of the BI tools with the new name.

PR #287, opened 2023-07-26
Update CEJST dataset to 1.0

bendnorman:
CEJST 1.0 adds a bunch of new columns and updates some underlying calculations. You can learn more about the changes from these resources:
* [Biden-Harris Administration Launches Version 1.0 of Climate and Economic Justice Screening Tool, Key Step in Implementing President Biden's Justice40 Initiative | CEQ | The White House](https://www.whitehouse.gov/ceq/news-updates/2022/11/22/biden-harris-administration-launches-version-1-0-of-climate-and-economic-justice-screening-tool-key-step-in-implementing-president-bidens-justice40-initiative/)
* [White House Releases version 1.0 of the Climate and Economic Justice Screening Tool \(CEJST\)](https://www.youtube.com/watch?v=XwilQp3EXRQ)
* [1.0-cejst-technical-support-document](https://static-data-screeningtool.geoplatform.gov/data-versions/1.0/data/score/downloadable/1.0-cejst-technical-support-document.pdf)

This PR updates the extract and transform code to handle the 1.0 data.  30% of the `counties_wide_format.justice40_dbcp_index` values changed between the beta and 1.0 update. Of the 30% of counties with new values, 55% of the values increased and 45% decreased.


src/dbcp/transform/justice40.py
TrentonBush, line 1: yeesh, thanks for renaming 689052734 columns!

src/dbcp/etl.py
```diff
@@ -217,6 +217,7 @@ def etl(args):
                 if_exists="append",
                 index=False,
                 schema="data_warehouse",
+                chunksize=1000,
```
TrentonBush, line 220: Did that help performance at all? Not the biggest deal on an ETL process that only takes 3 minutes, but I've always been shocked at how slow df.to_sql() is, no matter what kwargs I use. For my other work I had to make a backend-specific writer that used the `COPY x FROM STDIN WITH CSV` statement. It was literally 1000x faster. I should take the 10 minutes to copy it over here

.pre-commit-config.yaml
```diff
@@ -12,7 +12,7 @@ repos:
     rev: v4.0.1
     hooks:
       - id: check-added-large-files # Don't accidentally commit giant files.
-        args: ["--maxkb=10000"]
+        args: ["--maxkb=100000"]
```
TrentonBush, line 15: What do you think the cutoff should be for git vs GCS?

notebooks/31-bdn-justic40-update-analysis.ipynb
TrentonBush, line 1: I agree that the index differences are the main thing to look at here. I think the one missing piece here is a choropleth map of the differences. I think you can do that with geopandas, and [a county shapefile](https://www.census.gov/geographies/mapping-files/time-series/geo/carto-boundary-file.2014.html#list-tab-1556094155). One of the [diverging color maps](https://matplotlib.org/stable/tutorials/colors/colormaps.html#diverging) seems most appropriate.

We could do more, like looking for what specific index members drive the differences, but I''m not sure how deep they want to go. Or if Yujia wants to do it? I'll ask on slack.

TrentonBush (changes requested):
Nice! I think it is a map away from being ready to hand over

src/dbcp/etl.py
bendnorman: Yeah, I have had the same experience with `df.to_sql()`. I'm not sure if it made an improvement tbh.  I could do some more thorough testing if you like.

.pre-commit-config.yaml
bendnorman: GitHub's max file size is 100 MB, and Git LFS is 2 GB. I think it's simplest not to use the GCS bucket.

I don't think I was following this policy when I added the census zipfiles to the bucket. I could delete the GCS bucket in favor of Git LFS but it could be nice to keep around if we ever need a dataset that is larger that 2GB (aka PUDL).

notebooks/31-bdn-justic40-update-analysis.ipynb
bendnorman: Sounds good. I'll take a stab at a choropleth map. Given that most of the counties' indexes have stayed the same, I'm assuming a super in-depth analysis isn't unnecessary.

bendnorman:
Maps!
image

image


image

Not super legible without interacting with it in Jupyter. Here is some more legible information:
- There are 22 counties whose index was non-zero in the beta version and zero in 1.0.
- There are 200 counties whose index was 0 in the beta version and non zero in 1.0.

TrentonBush:
It would probably be more legible as raw differences than as percent. It's only a 6 point scale, IIRC

TrentonBush (approved):

PR #288, opened 2023-08-02
Analyze which counties fell above and below CEJST index 4.0 threshold

bendnorman:
This PR maps out which counties' CEJST index fell above and below the 4.0 threshold after the 1.0 update.

- 41 counties had a CEJST index < 4.0 in beta and >= 4.0 in 1.0
- 70 counties had a CEJST index > 4.0 in beta and < 4.0 in 1.0

Interesting that the mean CEJST index was higher in 1.0 than beta, but there was a net loss in counties with an index > 4.0. Here are the supporting maps:

image

image


TrentonBush (approved):
Looks good! If we do something like that in the future, I'd try to put them on the same map with something like white==unchanged, red==dropped out, blue==added. But I think this is perfectly legible. Also love to see 2.3 million lines deleted 😂 Embedded interactive plots are wild

bendnorman:
Yeah didn't realize how big they are! Had to clear the outputs because the notebook exceeded github's limit.

PR #294, opened 2023-09-10
Normalize ballot ready and create county commissioner data mart

bendnorman:
This PR:
- Normalizes the raw Ballot Ready data into three entity tables: races, elections, and positions.
- Reconstructs the denormalized `br_election_data` data mart from the normalized tables.
- Changes `county_commission_election_info` to have a row for each county and columns for the next general, special and primary elections.


TrentonBush:
I think the current normalization drops the m:m county:race relationship down to 1:m, and also assigns it as county:race when it is actually a simpler county:position relationship. So basically I think there should be an additional table: election level, position level, race level (aka election:position m:m table) and the new county:position table. Could you take a look at the notebook 36 I added and see if that makes sense to you?

bendnorman:
Ah good catch thank you! I moved the county information from the races table to a new county:position m:m table called `br_positions_counties`. I don't think this modeling change actually changes the shape of `data_mart.br_election_data`.

TrentonBush (approved):
k I just renamed the association table to be consistent with other m:m tables in the warehouse

PR #296, opened 2023-09-22
Gridstatus iso queues extraction and exploration

bendnorman:
This PR creates a function for caching and extracting grid status interconnection queue data.

I also upgraded:
- pandas-gbq~=0.19.1
- pydata-google-auth~=1.7.0


bendnorman:
@TrentonBush, `notebooks/37-bdn-explore-gridstatus-data.ipynb` is a good starting place for interacting with the archived gridstatus data. 

The archives are somewhat up to date. The combination of flaky ISO APIs and gridstatus basic pandas cleaning becoming out of date make the archiving a little flaky :/

bendnorman:
@TrentonBush I did some basic cleaning of the grid status data. A few of thoughts:
- Most of the columns in the LBNL data are shared amongst all of the ISOs in grid status so my cleaning focused on those columns. The notable exception is the detailed status information. it seems like each ISO reports this information in its own way so we'll probably need create separate cleaning functions for each iso instead of cleaning a concatenated version like I'm doing now. Would love your assistance for this part of the cleaning since you're more familiar with this!
- I mapped the technology types and would love for you to double check I made sensible decisions.
- Any idea why LBNL created one to many relationships between projects, capacity and location?
- To fully integrate GS with LBNL I'm imagining we'll create GS data warehouse tables, then adjust the county_wide / long data mart tables to pull non iso data from LBNL and ISO data from GS. What do you think?
- It looks like our LBNL code does some manual cleaning and deduplication. How can we determine if this need to be applied to the GS data?

src/dbcp/transform/gridstatus.py
```diff
@@ -0,0 +1,632 @@
...
+    active_projects = add_county_fips_with_backup_geocoding(
+        active_projects, state_col="state", locality_col="county"
+    )
+    # correct some fips codes
+    active_projects.loc[
+        active_projects.county_id_fips.eq("51515"), "county_id_fips"
+    ] = "51019"  # https://www.ddorn.net/data/FIPS_County_Code_Changes.pdf
+
+    # harmonize types
+    active_projects = _clean_resource_type(active_projects)
```
TrentonBush, line 623: in the LBNL data, hybrid projects were modeled as 1:m relationships between projects:resources. This is most obvious in CAISO, where there are columns like resource_1, resource_2 and capacity_1, capacity_2. But other ISOs that label them like "solar; storage" were modeled the same way. If they don't give separate capacity values, it was just duplicated.

src/dbcp/transform/gridstatus.py
```diff
@@ -0,0 +1,632 @@
...
+        "Transmission Owner": "transmission_owner",
+        "Winter Capacity (MW)": "winter_capacity_mw",
+        "Withdrawal Comment": "withdrawal_comment",
+        "Withdrawn Date": "withdrawn_date",
+        "is_actionable": "is_actionable",
+        "is_nearly_certain": "is_nearly_certain",
+        "iso_region": "iso_region",
+    }
+
+    iso_cleaning_functions = {
```
TrentonBush, line 578: This is more of an enhancement than a requirement, but it'd be nice to capture fuel and tech type separately, like EIA does, for the ISOs that report that info separately. That info would improve our CO2 estimates because we wouldn't have to guess if a gas plant is CC or CT. There is a factor of 4 difference in CO2 production!

TrentonBush (changes requested):

TrentonBush:
While trying to write transformed results to a .parquet file, I ran into an issue of mixed types in the output.:
```
d = df.convert_dtypes().dtypes
obj_cols = d[d.eq('object')].index
for c in obj_cols:
    types = gs_dfs_dec17[c].apply(type).unique()
    print(f"{c}:\t\t{types}\n")
```
output:
```
proposed_completion_date:		[<class 'datetime.datetime'> <class 'pandas._libs.missing.NAType'>
 <class 'pandas._libs.tslibs.timestamps.Timestamp'>
 <class 'pandas._libs.tslibs.nattype.NaTType'>]

queue_date:		[<class 'datetime.datetime'>
 <class 'pandas._libs.tslibs.timestamps.Timestamp'>
 <class 'pandas._libs.tslibs.nattype.NaTType'>]

queue_id:		[<class 'str'> <class 'int'>]

summer_capacity_mw:		[<class 'float'> <class 'pandas._libs.missing.NAType'> <class 'str'>]

winter_capacity_mw:		[<class 'float'> <class 'pandas._libs.missing.NAType'> <class 'str'>]

withdrawal_comment:		[<class 'pandas._libs.missing.NAType'>]```

File: src/dbcp/transform/gridstatus.py (hunk 1 above)
Comment (bendnorman) [reply src/dbcp/transform/gridstatus.py:623]: Yeah I didn't get around to sorting out the 1:m relationship in CAISO. I think this iteration of integration just grabs the `*_1` values for CAISO projects.

**Comment by bendnorman** [comment]
Just pushed a change that should correct the mixed dtype issue.

**Comment by bendnorman** [comment]
Ok! I realized by filtering by `"Status" == "Active"` I was removing a lot of projects that LBNL considered "active" and "nearly certain". Now I have some logic in each ISO cleaning function to filter out in service and withdrawn projects. I also normalized the data into projects and capacities. @TrentonBush your `normalize_multicolumns_to_rows` function was life saver thank you!

With these changes the overall capacity differences look pretty good for each ISO. I think the variability can be explained by the 12 month delay for SPP and ISONE and the 4 month delay for all other ISOs. There are a couple of remaining issues that I uncovered using notebook 42:

- Overall capacities look good for all the ISOs
- Same number of ISONE projects are marked actionable in old and new data but the MW distributions don’t line up.
- Significantly more projects marked actionable in new CAISO and MISO data
- Significantly less number of projects marked nearly certain in new MISO data.

File: src/dbcp/extract/gridstatus_isoqueues.py [hunk 3]
```diff
@@ -16,26 +16,27 @@
...
-    "nyiso": "1701730381901584",
+    "nyiso": "1681775159356063",
     "isone": "1701730382409516",
 }
 
 
 def extract(iso_queue_versions: dict[str, str] = ISO_QUEUE_VERSIONS):
     """Extract gridstatus ISO Queue data."""
     iso_queues: dict[str, pd.DataFrame] = {}
-    credentials = dbcp.extract.helpers.get_gcp_credentials()
```
TrentonBush: Is there a reason you changed this to get credentials inside the `for` loop, once for each queue?

README.md
```diff
@@ -43,6 +43,41 @@ conda activate dbcp-dev
 
 This conda environment has python, pip and pre-commit installed in it. This env is just for running pre-commits, the actual ETL development happens in docker.
 
+## GCP Authentication
```
bendnorman, line 46: Got fed up with authenticating everytime we ran the ETL so I finally figured out how to mount local user credentials into the docker container.

src/dbcp/data_mart/projects.py
```diff
@@ -15,20 +15,106 @@
...
+                    entity,
+                    utility,
+                    proposed_completion_date AS date_proposed_online,
+                    point_of_interconnection,
+                    UPPER(region) AS iso_region,
+                    is_actionable,
+                    resource_clean,
+                    queue_status,
+                    queue_date AS date_entered_queue,
+                    interconnection_status_raw AS interconnection_status
```
bendnorman, line 52: I'm using `interconnection_status_raw` for now because we map the raw data directly to nearly certain and actionable statuses.

src/dbcp/data_mart/projects.py
```diff
@@ -15,20 +15,106 @@
...
+
+    Returns:
+        A dataframe of ISO projects with location, capacity, estimated co2 emissions and state permitting info.
     """
     query = """
     WITH
     iso_proj_res as (
         SELECT
             proj.project_id,
+            proj.queue_id,
```
bendnorman, line 117: I added `queue_id` because it is helpful when exploring and comparing the `iso_projects_long_format` table.

src/dbcp/data_mart/projects.py
```diff
@@ -349,6 +400,7 @@ def _add_derived_columns(mart: pd.DataFrame) -> None:
         "Solar; Storage": "renewable",
         "Solar": "renewable",
         "Steam": np.nan,
+        "Transmission": "transmission",
```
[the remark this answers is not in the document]
bendnorman: Agreed! Would be good to check in. I added a task to the GS issue.

src/dbcp/extract/gridstatus_isoqueues.py
```diff
@@ -0,0 +1,44 @@
...
+ISO_QUEUE_VERSIONS: dict[str, str] = {
+    "miso": "1681775160487863",
+    "caiso": "1681775162586588",
+    "pjm": "1681775160979859",
+    "ercot": "1681775161342766",
+    "spp": "1701730381410448",
+    "nyiso": "1681775159356063",
+    "isone": "1701730382409516",
+}
+
```
bendnorman, line 31: This needs to be updated to use the latest versions.

src/dbcp/validation/tests.py
```diff
@@ -270,14 +282,16 @@ def validate_warehouse(engine: Engine):
...
 
 
 def validate_data_mart(engine: Engine):
     """Run data mart validation tests."""
     logger.info("Validating data mart")
     test_county_long_vs_wide(engine)
     test_county_wide_coverage(engine)
-    test_iso_projects_data_mart_aggregates_are_close(engine)
+    # TODO (bendnorman): This is currently failing but a small percentage
+    # test_iso_projects_data_mart_aggregates_are_close(engine)
```
bendnorman, line 294: @TrentonBush I don't really understand what's going on in this test.

TrentonBush:
OK I finally added the deduplication and fixed the output tests. I think this is ready to go?

PR #297, opened 2023-10-05
Use copy CSV instead of insert for pandas to postgres

bendnorman:
This PR:
- Improve performance by chaniging the `method` argument of `to_sql` calls to use a function [from the pandas docs](https://pandas.pydata.org/docs/user_guide/io.html#io-sql-method) that uses COPY CSV instead of INSERT.
- Creates a helper function that enforces datatypes on dataframes based on the types specified in the metadata.
- Changes `data_mart.fossil_infrastructure_projects.cost_millions` column from an Int to Float because I was getting a "could not cast int to float" error when enforcing the types.

On my comp `make etl_local` went from 6m39s -> 2m58s.

### DIFF
--- a/src/dbcp/data_mart/__init__.py
+++ b/src/dbcp/data_mart/__init__.py
@@ -7,6 +7,7 @@ import pkgutil
 import pandas as pd
 
 import dbcp
+from dbcp.helpers import enforce_dtypes, psql_insert_copy
 from dbcp.metadata.data_mart import metadata
 from dbcp.validation.tests import validate_data_mart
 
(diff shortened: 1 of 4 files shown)


TrentonBush:
Does this work if the df columns are in a different order than the metadata definition? Just noticed the csv is written separately from the COPY statement that defines the headers

bendnorman:
Good question! I tested out mixing up the order of tables, and `df.to_sql()` still inserted the data with the correct column order.

TrentonBush (approved):

PR #299, opened 2023-12-09
add manual airtable ordinances

TrentonBush:
Add a new warehouse table with manually updated ordinance data pulled from BigQuery (originally in an Airtable).
* add one new column (`ordinance_via_self_maintained`) to the county and ISO project data mart tables
* redefine the `ordinance_is_restrictive` column to prioritize this data source
* update the docs accordingly

I added one validation test to check that the incoming data exactly matches the county FIPS codes from the Census (in table `data_warehouse.county_fips`).


bendnorman (approved):
LGTM! We don't need to change anything in the c4 repo, correct? Also, I updated airtable to just pull `county_id_fips` and `ordinance_status`.

TrentonBush:
> We don't need to change anything in the c4 repo, correct?

Barely, just update a couple of SELECT queries. I made a PR there the other day

PR #304, opened 2024-01-17
Update october 2023 ballot ready data

bendnorman:
This PR updates the ballot ready data to use the October 2023 data. There are a few things to note:

- Based on their data dictionary`frequency` and `reference_year` are supposed to be required but there are a 345 nulls in this update.
- There are about 25 weird edge cases positions (0.04%) that have multiple frequencies and reference years because ballot ready keep the old and new values if there is a redistricting or law change.
- There are about 3x as many races in this batch. I assume because it is an election year.
- There are some races that still use the Valdez-Cordova Census Area. I copied all of the races between the Copper River and Chugach census areas.

### DIFF
--- a/src/dbcp/transform/ballot_ready.py
+++ b/src/dbcp/transform/ballot_ready.py
@@ -42,8 +42,6 @@ def _normalize_entities(ballot_ready: pd.DataFrame) -> dict[str, pd.DataFrame]:
     trns_dfs["br_elections"] = br_elections
 
     # Positions
-    position_pk_fields = ["position_id"]
-
     position_fields = [
         "reference_year",
         "position_id",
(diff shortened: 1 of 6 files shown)


src/dbcp/transform/ballot_ready.py
```diff
@@ -57,19 +61,44 @@ def _normalize_entities(ballot_ready: pd.DataFrame) -> dict[str, pd.DataFrame]:
...
+            new_index, new_index + n_duplicates
+        )
+        new_position_id_mapping = duplicate_position_ids.set_index(
+            "position_id"
+        ).to_dict()["new_position_id"]
+
+        ballot_ready["position_id"] = ballot_ready["position_id"].replace(
+            new_position_id_mapping
+        )
+
```
bendnorman, line 95: This doesn't feel great to me. I think a better option might be to break out the one to many relationship between `position_id` and `frequency` and `reference_year` into another table. What do you think @TrentonBush

src/dbcp/transform/ballot_ready.py
```diff
@@ -155,11 +184,12 @@ def _explode_counties(raw_ballot_ready: pd.DataFrame) -> pd.DataFrame:
...
-    # All elections are state and federal level so I will duplicate the races for the two new census areas
+    # It is reasonable to split State and Federal elections between the two areas.
+    # However, adding local elections to both counties is not appropriate. I'm going to do it
+    # anyways because there aren't any great options for accurately geocoding the local elections.
     valdez = ballot_ready.query("county_id_fips == '02261'")
-    assert valdez.level.isin(
-        ["state", "federal"]
-    ).all(), "Found a local election in the Valdez-Cordova Census Area!"
+    if valdez.level.isin(["state", "federal"]).all():
+        logger.info("Found a local election in the Valdez-Cordova Census Area!")
```
bendnorman, line 192: I figured it's better to have duplicates than to lose these elections or do a bad job at geocoding them.

TrentonBush (changes requested):
I think we can drop the duplicate positions without much impact if we use not-null and election_day to pick the most relevant values. See my longer in-line comment for details.

bendnorman:
That all makes sense! Just updated the deuplication logic.

src/dbcp/transform/ballot_ready.py
```diff
@@ -67,35 +65,29 @@ def _normalize_entities(ballot_ready: pd.DataFrame) -> dict[str, pd.DataFrame]:
...
+        len(duplicate_positions) <= 52
+    ), f"Found more duplicate positions than expected: {len(duplicate_positions)}"
+
+    # dropnas in frequency and reference_year
+    duplicate_positions = duplicate_positions.dropna(
+        subset=["frequency", "reference_year"]
+    )
+    # select the the position that has the most recent election day
+    latest_position_idx = duplicate_positions.groupby("position_id")[
+        "election_day"
```
TrentonBush, line 82: Does this run? Might need to change the `br_positions` creation line to include `subset=position_fields + ['election_day']`

bendnorman: This does run! I don't think the final positions table should actually include `election_day`.

src/dbcp/transform/ballot_ready.py
```diff
@@ -65,7 +67,7 @@ def _normalize_entities(ballot_ready: pd.DataFrame) -> dict[str, pd.DataFrame]:
     # and the future/current one are reflected in the data.
 
     # check if position_id is unique
-    br_positions = ballot_ready.drop_duplicates(subset=position_fields)
+    br_positions = ballot_ready.drop_duplicates(subset=position_pk_fields)
     is_duplciate_position = br_positions.position_id.duplicated(keep=False)
```
TrentonBush, line 71: I don't understand this change. We drop duplicates based on `position_id` then look for duplicates of `position_id`?

bendnorman: Oh i'm a ding dong. Thanks for the catch. I changed it to drop using all position columns.

TrentonBush (approved):

PR #307, opened 2024-02-05
Update PUDL ETL to pull new data directly from database

bendnorman:
Still getting mysterious error that 'battery_storage_existing_co2e_tonnes_per_year' does not exists in a data mart dataframe. Not sure why a PUDL update would cause this.


TrentonBush:
PUDL may have renamed the battery storage label or they were filtered out from MCOE? That column is created basically like a pivot table, so if there are no battery storage values maybe the column doesn't get created. I can follow up in this upcoming sprint if you'd like.

bendnorman:
On `dev`, `battery_storage_existing_co2e_tonnes_per_year` only had one non null county. The updated pudl data is all nulls so the column is dropped during [this step](https://github.com/deployment-gap-model-education-fund/deployment-gap-model/blob/b5b08b025af55904899bcbb4d25209df0e897de2/src/dbcp/data_mart/counties.py#L585) so I removed it from the list of columns to drop.

With the postgres load improvement and loading PUDL tables from the DB, the full ETL takes 3 min to run on my computer.

src/dbcp/data_mart/counties.py
```diff
@@ -181,7 +181,7 @@ def _get_existing_plant_attributes(engine: sa.engine.Engine) -> pd.DataFrame:
     #         plant_id_eia,
     #         generator_id,
     #         count(fuel_type_code_pudl) as fuel_type_count
-    #     FROM "data_warehouse"."mcoe"
+    #     FROM "data_warehouse"."pudl_generators"
```
bendnorman, line 184: The table formerly known as `mcoe` was renamed in PUDL because it didn't fully compute MCOE and is really a table of generators and all it's attributes. I renamed the data warehouse table to be more consistent with what is actually in the table.

src/dbcp/extract/pudl.py
```diff
@@ -0,0 +1,40 @@
...
+    Args:
+        pudl_engine: The pudl sqlite database engine.
+
+    Returns:
+        The pudl_generators table.
+    """
+    with pudl_engine.connect() as con:
+        pudl_generators = pd.read_sql(
+            "SELECT * FROM out_eia__yearly_generators WHERE report_date >= '2022-01-01' AND report_date < '2023-01-01'",
+            con,
```
bendnorman, line 20: Update to use `PUDL_LATEST_YEAR`.

TrentonBush (changes requested):

PR #309, opened 2024-02-13
Init docker compose action

bendnorman:
This PR does many things:
- Creates a github action called `full-build` that runs the full ETL and tests every time a commit is pushed. yay CI!
- The ETL now writes the final dataframes to postgres and parquet files. The new `src/dbcp/publish.py` script uploads the parquet files to `gs://dgm-outputs` and loads the files from the bucket into BigQuery. The bucket is organized by data version and schema. `dev` will have the latest data from the `dev` branch, and `prod` will have the data from the latest pinned release. The data version is added as a label to the BQ tables so users know which version of the data they are using in BQ.
- `full-build` runs `publish.py` when a tag is pushed or when a new commit is pushed into `dev`. yay CD!
- The environment variables were kind of all over the place. We had `.env`, `local.env`, `default.env` and instructions to set local environment variables. I simplified this by adding all of the "static" env vars to .env and checking it into git. `.env` variables are available within the docker container and the `docker-compose.yml` file. Secrets should should not be tracked in git are pulled from the local environment. So now when a user sets up the repo they just need to set `GOOGLE_GHA_CREDS_PATH`, and `API_KEY_GOOGLE_MAPS` in their local environment.
- I removed the BQ upload logic in `dbcp.etl` and `dbcp.data_mart` now that `publish.py` handles the uploading. BQ should only be updated in the GitHub action.
- `pudl.sqlite` was too big for GitHub actions so I'm pulling PUDL data from `gs://parquet.catalyst.coop` as parquet files. 

With this PR we have: CI, CD, the ability to compare previous version of our data.


src/dbcp/transform/local_opposition.py
```diff
@@ -96,6 +96,9 @@ def _transform_local_ordinances(local_ord_df: pd.DataFrame) -> pd.DataFrame:
     # manual corrections
     location_corrections = {
         "Batavia Township (Clermont County)": "Batavia Township (Branch County)",
+        "Town of Albion (Kennebec County)": "Albion (Kennebec County)",
+        "Town of Lovell (Oxford County)": "Lovell (Oxford County)",
+        "Town of Charlton (Worcester County)": "Charlton (Worcester County)",
```
bendnorman, line 101: Got a null violation constraint in the CI because it had a fresh cache. I'm not sure why my older local cache was able to geocode these, but making these corrections worked.

TrentonBush:
Your note about the local cache reminded me that the point of the cache was to avoid blasting the API with thousands of queries every time I ran the tests. The Maps Platform API is [free for 40k queries per month and is a half cent per query after](https://developers.google.com/maps/documentation/geocoding/usage-and-billing). I dont remember how many queries happen in each run, but I think it was several thousand. We should keep an eye on this if the CI runs frequently.

bendnorman:
Ah good point. I should look into how many requests are made for an ETL run so the CI doesn't blow through our quota.

PR #312, opened 2024-03-12
Switch GS archive bucket

bendnorman:
The gridstatus archives are [now being saved](https://github.com/deployment-gap-model-education-fund/deployment-gap-model-archiver/pull/18) to `gs://dgm-archive` instead of having a separate bucket for just the GS data. This PR updates the extraction code in the ETL to use the new bucket.


src/dbcp/extract/gridstatus_isoqueues.py
```diff
@@ -20,21 +20,23 @@
...
-    "spp": "1704654954488739",
-    "nyiso": "1702235705611699",
-    "isone": "1704654954804863",
+    "miso": "1709776341526904",
+    "caiso": "1709776259365612",
+    "pjm": "1709776387308475",
+    "ercot": "1709776286037540",
+    "spp": "1709776417253050",
+    "nyiso": "1709776369660949",
+    "isone": "1709776310995150",
```
bendnorman, line 29: These are the same files just new revision numbers. Copying the parquet GS archives to the bucket generated new revision numbers. I linked the old and new revision numbers using the file hash generated by GCS.

PR #321, opened 2024-04-25
Add MISO projects prior to 2017 using old GS snapshot

bendnorman:
At some point in 2023 MISO removed all projects that entered the queue prior to 2017 so our newer archives of the data only have projects that entered after 2017. Lucikly our older archives include projects prior to 2017. This PR combines the newer archive with any projects in the old archive that don't exist in the new one.

### DIFF
--- a/src/dbcp/extract/gridstatus_isoqueues.py
+++ b/src/dbcp/extract/gridstatus_isoqueues.py
@@ -18,6 +18,7 @@ logger = logging.getLogger(__name__)
 
 ISO_QUEUE_VERSIONS: dict[str, str] = {
     "miso": "1712351594795115",
+    "miso-pre-2017": "1709776311574737",
     "caiso": "1712351595132954",
     "pjm": "1712351595509644",
     "ercot": "1712351595819850",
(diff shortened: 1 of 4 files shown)


src/dbcp/transform/gridstatus.py
```diff
@@ -591,8 +591,28 @@ def _create_project_status_classification_from_multiple_columns(
...
+    # grab projects that are only in pre_2017
+    only_in_pre_2017 = pre_2017[~pre_2017["queue_id"].isin(post_2017["queue_id"])]
+
+    # ensure there are no active projects in only_in_pre_2017
+    assert (
+        only_in_pre_2017["queue_status"].ne("Active").all()
+    ), "There are active projects in the pre-2017 MISO data that are not in the current MISO data."
+
+    # concat only_in_pre_2017 with iso_df
+    iso_df = pd.concat([post_2017, only_in_pre_2017])
```
TrentonBush, line 614: I saw that there are no column differences between the two datasets, but I thought there would be because of code changes in GridStatus since the old archive was created. For my own background context, why is that not the case? Did the archive happen after the GS changes? Was MISO not one of the affected ISOs? Or did something other than column names change?

TrentonBush (approved):
This extension looks good to me. I don't like that we have unhandled, erroneous, duplicate locations but that seemed out of scope for this PR and not a large enough error to be urgent. I made a separate issue #335 for that.

src/dbcp/extract/gridstatus_isoqueues.py
```diff
@@ -18,6 +18,7 @@
 
 ISO_QUEUE_VERSIONS: dict[str, str] = {
     "miso": "1712351594795115",
+    "miso-pre-2017": "1709776311574737",
```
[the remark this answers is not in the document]
bendnorman: Good question. Yeah the creation dates are when I migrated the data from the gridstatus-archive bucket to the dgm-archive bucket. Going forward, the creation dates will be accurate. I'm open to redesigning how we archive GS so we can track more metadata and it's more clear what data version you're working with.

src/dbcp/transform/gridstatus.py
bendnorman: It doesn't look like there have been any column name changes in the [last couple of years](https://github.com/kmax12/gridstatus/blame/main/gridstatus/miso.py).

TrentonBush: I mean isn't that old archive of processed GridStatus outputs, the code for which has changed since then? Or is that archive of raw data that is safe from those changes?

PR #329, opened 2024-04-30
Widen ISO queue change log

bendnorman:
This PR creates four new data mart tables:
- `iso_counties_new_clean_capacity_mw_changelog`
- `iso_counties_new_clean_n_projects_changelog`
- `iso_regions_new_clean_capacity_mw_changelog`
- `iso_regions_new_clean_n_projects_changelog`

Each row in these tables is a geography, columns are the year a month a quarter ended and the values are the number of new clean projects or capacity mw for the given geography and quarter.


TrentonBush (approved):
Looks good! I had to remind myself why we allow 2 duplicate projects in the long format table, so I added some comments about that and couple of asserts to check it for future changes. No changes to the tables you made.

PR #332, opened 2024-05-06
draft 860m projects tables. needs metadata and debug

TrentonBush:
This PR creates 3 data mart tables based on 860m data and their associated documentation.


src/dbcp/metadata/data_mart.py
```diff
@@ -537,3 +533,45 @@
...
 )
+
+
+projects_history_860m = Table(
+    "projects_history_860m",
+    metadata,
+    Column("plant_name_eia", String),
+    Column("plant_id_eia", Integer, primary_key=True),
+    Column("generator_id", String, primary_key=True),
+    Column("status_2020-12-31", String),
```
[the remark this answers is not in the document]
bendnorman: I feel like we should include the value description in the table name not the column name. As for the date, should we do YYYY-MM-DD, YYYY-MM or YYYY-Quarter?

src/dbcp/data_mart/projects.py
```diff
@@ -540,13 +540,172 @@ def create_long_format(engine: sa.engine.Engine) -> pd.DataFrame:
...
+    """Get the most recent EIA860M data."""
+    if not date_as_of:  # get most recent data
+        date_as_of = (
+            pd.read_sql(
+                "SELECT max(valid_until_date) FROM data_warehouse.pudl_eia860m_changelog",
+                engine,
+            )
+            .iat[0, 0]
+            .strftime("%Y-%m-%d")
+        )
```
bendnorman, line 555: Why do you like to use SQL for the data mart tables? I feel like we should move away from storing complex SQL as python strings because it's difficult to format and read. We're also better at pandas than SQL.

src/dbcp/data_mart/projects.py
```diff
@@ -540,13 +540,172 @@ def create_long_format(engine: sa.engine.Engine) -> pd.DataFrame:
...
 
-def _pudl_eia860m_changelog(engine: sa.engine.Engine) -> pd.DataFrame:
-    """Get the PUDL EIA860M changelog table."""
-    with engine.connect() as con:
-        pudl_eia860m_changelog = pd.read_sql_table(
-            "pudl_eia860m_changelog", con, schema="data_warehouse"
+def get_eia860m_current(
+    engine: sa.engine.Engine, date_as_of: Optional[str] = None
+) -> pd.DataFrame:
+    """Get the most recent EIA860M data."""
```
bendnorman, line 546: Can you add argument and return value descriptions to the doc strings please.

src/dbcp/data_mart/projects.py
```diff
@@ -540,13 +540,172 @@ def create_long_format(engine: sa.engine.Engine) -> pd.DataFrame:
...
         )
-    return pudl_eia860m_changelog
+    status_history["status_current"] = status_history["operational_status_code"].where(
+        status_history["end_date"] == end_date
+    )
+    out = status_history.groupby(["plant_id_eia", "generator_id"], as_index=False)[
+        [c for c in status_history.columns if c.startswith("status_")]
+    ].first()
+    eia860m_plant_names = _get_plant_names(engine)
+    out = out.merge(eia860m_plant_names, on="plant_id_eia", how="left")
```
bendnorman, line 649: Could this table be constructed more easily using `pivot()`?

bendnorman (changes requested):

src/dbcp/metadata/data_mart.py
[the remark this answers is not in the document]
TrentonBush: I totally agree! Sorry I forgot to align these earlier. Also I don't like having dynamic column names in a database, period. But it was a direct request. I think we should push them to restructure it to long format if they can make it work on their end, while it is still in dev and things aren't built on top of it yet.

src/dbcp/data_mart/projects.py
TrentonBush: I like your point about our organizational competence being in pandas. You're probably right that most things should be in pandas vs SQL. I started using SQL here for a couple of reasons, but none are super compelling for this use case:
 
* use what users use. The final product is a SQL database. I wanted to interact with the data in the same way they would so I could be more helpful and also discover pain points first hand. But it turns out our users mostly don't use SQL so I don't think this matters here.
* development convenience. Because the data warehouse lives in a SQL database, it is faster to jump in with SQL, so I do most prototyping in SQL. I find SQL is more succinct for a lot of basic operations like subsetting, grouping, and joining. But as I got more comfortable with SQL my definition of "basic" has expanded, and probably too far.
* performance. It is more efficient to subset a table at the database rather than transferring the whole thing to pandas just to throw most of it away. But performance doesn't matter on this small data -- this is more about engineer's ego than real value 😁

Looking at these particular 860m queries, they do still mostly seem to me to be basic SELECT operations, with the possible exception of the less well known COALESCE() function.  I definitely agree that some of the longer queries should have been in pandas (like `_get_proprietary_proposed_offshore()`). But it sounds like my judgement there doesn't match yours. Where do you think we should draw the line? Should the only interaction with the database be to `pd.read_sql_table()`, not to run any queries?

src/dbcp/data_mart/projects.py
TrentonBush: Yes, will do. I forgot to document those because I didn't end up implementing the logic. Maybe I'll just remove it for now.

src/dbcp/data_mart/projects.py
bendnorman: All valid points! Yeah, given the data is pretty small, reading the whole table in using `pd.read_sql_table()` feels reasonable.

I think another reason to rely on mostly pandas is it's backend agnostic. I tried to convert parts of the ETL to use SQLite and ran into issues because there are some postgres specific functions in the queries.

PR #334, opened 2024-05-17
Reshape ISO queue change logs

bendnorman:
This PR
- Removes the wide versions of the change log which had dates in the columns which are annoying to maintain because we'll have to update the column names each quarter.
- Creates 4 new data mart tables that are snapshots of active projects each quarter for a given geography. This allows users to see the total active capacity for previous years. 
- Renames `iso_regions_change_log` and `iso_counties_change_log` to `iso_regions_all_projects_change_log` and `iso_regions_all_projects_change_log`. These tables give users a sense of projects that entered and exited the queue for a given time period and geography. The four new tables just show the total active projects for previous years.


src/dbcp/metadata/data_mart.py
```diff
@@ -529,64 +527,46 @@
     schema=schema,
 )
 
-iso_counties_new_clean_n_projects_changelog = Table(
-    "iso_counties_new_clean_n_projects_changelog",
+counties_active_projects_capacity_mw_change_log = Table(
```
TrentonBush, line 530: Could these 4 tables be condensed to 2? I see two tables with PK (`county_id_fips`, `report_date`) and two with PK (`iso_region`, `report_date`). Or was this format a specific request?

src/dbcp/metadata/data_mart.py
```diff
@@ -529,64 +527,46 @@
...
+    Column("report_date", DateTime, primary_key=True),
+    Column("clean_capacity_mw", Float),
+    Column("fossil_capacity_mw", Float),
+    Column("other_capacity_mw", Float),
     schema=schema,
 )
-iso_counties_new_clean_capacity_mw_changelog = Table(
-    "iso_counties_new_clean_capacity_mw_changelog",
+
+iso_regions_active_projects_capacity_mw_change_log = Table(
```
TrentonBush, line 541: Do these tables need documentation or are they self-explanatory?

src/dbcp/metadata/data_mart.py
bendnorman: I proposed combining the tables but they requested four separate tables.

src/dbcp/metadata/data_mart.py
bendnorman: Whoops! I totally wrote documentation but I forgot to push it!

PR #340, opened 2024-06-11
Reshape changelogs

TrentonBush:
Make a new PR for #334 because I erroneously merged it and had to revert.


bendnorman:
Ok, I think the issue was that I was using `project_id` as part of the natural key instead of the `surrogate_id`.

PR #345, opened 2024-06-24
Update offshore wind data

bendnorman:
This PR updates the proprietary offshore wind data. To archive the raw data I just downloaded it directly from airtable and dropped it into the `gs://dgm-archive`. The modeling is mostly the same as the original data but I added an association table for port locations. There are a few other one to many relationships between projects and other attributes by decided to just focus on location columns for now.


src/dbcp/data_mart/projects.py
```diff
@@ -274,7 +275,7 @@ def _get_proprietary_proposed_offshore(engine: sa.engine.Engine) -> pd.DataFrame
     )
     -- join the project, state, and county stuff
     SELECT
-        assoc.project_id,
+        proj.project_id,
```
bendnorman, line 278: Projects that don't have a location were being dropped because of the `INNER JOIN` and pulling the `project_id`s from the association table.

src/dbcp/data_mart/projects.py
```diff
@@ -299,15 +300,15 @@ def _get_proprietary_proposed_offshore(engine: sa.engine.Engine) -> pd.DataFrame
...
+    RIGHT JOIN data_warehouse.offshore_wind_projects as proj
     USING(project_id)
     LEFT JOIN data_warehouse.state_fips as sfip
     ON substr(assoc.county_id_fips, 1, 2) = sfip.state_id_fips
     LEFT JOIN data_warehouse.county_fips as cfip
     USING(county_id_fips)
     LEFT JOIN data_warehouse.ncsl_state_permitting as ncsl
     ON substr(assoc.county_id_fips, 1, 2) = ncsl.state_id_fips
-    WHERE proj.construction_status != 'Online'
+    WHERE proj.construction_status != 'Online' OR proj.construction_status IS NULL
```
bendnorman, line 311: Some projects have missing construction status.

TrentonBush:
Also how did you export the raw CSVs? Did you use somebody else's Airtable view? I'd consider making our own like "Catalyst Export" so other people's changes don't impact us. Plus we can prep it how we want to make normalization easier, if needed.

data/raw/offshore_wind_locations.csv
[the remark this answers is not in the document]
bendnorman: These csvs are from the old table which is why I deleted them. The raw locations and projects csvs stored in [`gs://dgm-archive`](https://console.cloud.google.com/storage/browser/dgm-archive/synapse/offshore_wind;tab=objects?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&project=dbcp-dev-350818&prefix=&forceOnObjectsSortingFiltering=false) match the airtable views.

src/dbcp/metadata/data_warehouse.py
```diff
@@ -1175,15 +1189,52 @@
...
+        primary_key=True,
+    ),
+    Column(
+        "project_id",
+        Integer,
+        ForeignKey("data_warehouse.offshore_wind_projects.project_id"),
+        primary_key=True,
+    ),
     schema=schema,
 )
```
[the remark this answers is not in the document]
bendnorman: Ah thank you! That's what a get for copy pasting code :/

bendnorman:
I just downloaded a csv from the "Cleaned View". It seems like data between views stay synced the the columns and layout your choose are specific to the view. So is the benefit of creating our own view that the columns will stay the same between updates? In other words, if they change the "cleaned view" view layout, the "catalyst export" view's layout will stay the same?

data/raw/offshore_wind_locations.csv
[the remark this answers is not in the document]
TrentonBush: Oh oops I saw them in the change list and assumed they were added. Now it makes sense!

TrentonBush:
> is the benefit of creating our own view that the columns will stay the same between updates?
Yep, exactly

src/dbcp/transform/offshore_wind.py
```diff
@@ -96,56 +173,6 @@ def _add_geocoded_locations(transformed_locs: pd.DataFrame) -> None:
...
-    ports_to_locations = pd.MultiIndex.from_frame(
-        _id_association_table_from_csv_array(
-            proj["Port Location IDs"], name="location_id"
-        )[["location_id", "project_id"]]
-    )
-    assert locations_to_ports.symmetric_difference(ports_to_locations).empty
-    return
-
-
 def _add_actionable_and_nearly_certain_classification(projects: pd.DataFrame) -> None:
```
TrentonBush, line 176: This code was unchanged but I feel like this should be based on `overall_project_status` rather than `construction_status`. I'll check with Yujia

TrentonBush: K she said to keep it

TrentonBush (approved):

PR #349, opened 2024-07-02
Offshore tweaks

TrentonBush:
Offshore wind projects were defined differently from LBNL and Gridstatus projects. There was a county level groupby in the `src/dbcp/data_mart/projects.py:_get_proprietary_proposed_offshore` definition. By removing this, we treat multi-location counties consistently between all three sources. This also lets us remove per-dataset special testing from `src/dbcp/validation/tests.py:test_iso_projects_capacity_aggs`.

I also added a `queue_status` equivalent so the "active projects only" filter works as intended.

### DIFF
--- a/src/dbcp/data_mart/projects.py
+++ b/src/dbcp/data_mart/projects.py
@@ -253,25 +253,19 @@ def _get_proprietary_proposed_offshore(engine: sa.engine.Engine) -> pd.DataFrame
     """
     query = """
     WITH
-    cable_locs as (
-        SELECT
-            project_id,
-            locs.county_id_fips,
-            COUNT(*) OVER(PARTITION BY project_id) AS n_locations
-        FROM data_warehouse.offshore_wind_cable_landing_association as cable
-        INNER JOIN data_warehouse.offshore_wind_locations as locs
-        USING(location_id)
-    ),
     proj_county_assoc as (
         SELECT
             project_id,
-            county_id_fips,
-            -- some counties have multiple cable landings from the same
-            -- project (different towns). I allocate the capacity equally
-            -- over the landings
-            (count(*) * 1.0 / max(n_locations))::real as frac_locations_in_county
-        FROM cable_locs
-        group by 1,2
+            locs.county_id_fips,
+            -- Note that "frac_locations_in_county" is a misnomer. It is actually
+            -- "fraction_of_locations_represented_by_this_row". When I originally
+            -- named it, I thought location:county was m:1, but it's actually m:m
+            -- because some projects list multiple towns in the same parent county
+            -- in the raw "county" field.
+            (1.0 / count(*) over (partition by project_id))::real as frac_locations_in_county
+        FROM data_warehouse.offshore_wind_cable_landing_association as cable
+        INNER JOIN data_warehouse.offshore_wind_locations as locs
+        USING(location_id)
     )
     -- join the project, state, and county stuff
     SELECT
@@ -288,7 +282,16 @@ def _get_proprietary_proposed_offshore(engine: sa.engine.Engine) -> pd.DataFrame
         proj.developer,
         proj."capacity_mw",
         date(proj.proposed_completion_year::text || '-01-01') as date_proposed_online,
-        'active' as queue_status,
+        CASE coalesce(proj.construction_status, 'TBD') -- map to ISO queue_status categories
+            WHEN 'Online' THEN 'completed'
+            WHEN 'Canceled' THEN 'withdrawn'
+            WHEN 'Suspended' THEN 'withdrawn'
+            WHEN 'Not started' THEN 'active'
+            WHEN 'Construction underway' THEN 'active'
+            WHEN 'Site assessment underway' THEN 'active'
+            WHEN 'TBD' THEN 'active' -- assume active
+            ELSE NULL -- shouldn't happen
+            END as queue_status,
         'Offshore Wind' as resource_clean,
         0.0 as co2e_tonnes_per_year,
         proj.is_actionable,
@@ -299,8 +302,8 @@ def _get_proprietary_proposed_offshore(engine: sa.engine.Engine) -> pd.DataFrame
         cfip.county_name as county,
         ncsl.permitting_type as state_permitting_type
 
-    FROM proj_county_assoc as assoc
-    RIGHT JOIN data_warehouse.offshore_wind_projects as proj
+    FROM data_warehouse.offshore_wind_projects as proj
+    LEFT JOIN proj_county_assoc as assoc
     USING(project_id)
     LEFT JOIN data_warehouse.state_fips as sfip
     ON substr(assoc.county_id_fips, 1, 2) = sfip.state_id_fips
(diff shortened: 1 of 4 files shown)


src/dbcp/data_mart/projects.py
```diff
@@ -253,25 +245,19 @@ def _get_proprietary_proposed_offshore(engine: sa.engine.Engine) -> pd.DataFrame
     """
     query = """
     WITH
-    cable_locs as (
+    proj_county_assoc as (
```
TrentonBush, line 248: Removing the project-county group by lets us combine the two CTEs into one

src/dbcp/data_mart/projects.py
```diff
@@ -299,19 +294,21 @@ def _get_proprietary_proposed_offshore(engine: sa.engine.Engine) -> pd.DataFrame
         cfip.county_name as county,
         ncsl.permitting_type as state_permitting_type
 
-    FROM proj_county_assoc as assoc
-    RIGHT JOIN data_warehouse.offshore_wind_projects as proj
+    FROM data_warehouse.offshore_wind_projects as proj
```
TrentonBush, line 297: swap left/right for consistency. makes it easier to read

src/dbcp/data_mart/projects.py
```diff
@@ -299,19 +294,21 @@ def _get_proprietary_proposed_offshore(engine: sa.engine.Engine) -> pd.DataFrame
...
     ON substr(assoc.county_id_fips, 1, 2) = sfip.state_id_fips
     LEFT JOIN data_warehouse.county_fips as cfip
     USING(county_id_fips)
     LEFT JOIN data_warehouse.ncsl_state_permitting as ncsl
     ON substr(assoc.county_id_fips, 1, 2) = ncsl.state_id_fips
-    WHERE proj.construction_status != 'Online' OR proj.construction_status IS NULL
     ;
     """
     df = pd.read_sql(query, engine)
+    assert (
```
TrentonBush, line 309: protect against new statuses in updates

src/dbcp/data_mart/projects.py
```diff
@@ -569,14 +566,6 @@ def create_long_format(
         combined_opp, on="county_id_fips", how="left", validate="m:1"
     )
     _add_derived_columns(long_format)
-    pk = ["source", "project_id", "county_id_fips", "resource_clean"]
```
TrentonBush, line 572: Allow multi-location counties, which are the source all ten duplicates.

src/dbcp/validation/tests.py
```diff
@@ -196,25 +197,6 @@ def test_iso_projects_capacity_aggs(engine: Engine):
     absolute_diff = data_mart - source
     relative_diff = absolute_diff / source
 
-    # Some projects in offshore wind data have multiple locations, so the capacity is counted multiople times during the aggregation
```
TrentonBush, line 199: no more exceptions

src/dbcp/data_mart/projects.py
```diff
@@ -288,7 +274,16 @@ def _get_proprietary_proposed_offshore(engine: sa.engine.Engine) -> pd.DataFrame
...
+        CASE coalesce(proj.construction_status, 'TBD') -- map to ISO queue_status categories
+            WHEN 'Online' THEN 'completed'
+            WHEN 'Canceled' THEN 'withdrawn'
+            WHEN 'Suspended' THEN 'withdrawn'
+            WHEN 'Not started' THEN 'active'
+            WHEN 'Construction underway' THEN 'active'
+            WHEN 'Site assessment underway' THEN 'active'
+            WHEN 'TBD' THEN 'active' -- assume active
+            ELSE NULL -- shouldn't happen
+            END as queue_status,
```
bendnorman, line 286: Should we do this in dbcp.transform.offshore_wind? Have a queue_status and raw_queue_status column?

src/dbcp/data_mart/projects.py
```diff
@@ -288,7 +274,16 @@ def _get_proprietary_proposed_offshore(engine: sa.engine.Engine) -> pd.DataFrame
         proj.developer,
         proj."capacity_mw",
         date(proj.proposed_completion_year::text || '-01-01') as date_proposed_online,
-        'active' as queue_status,
+        CASE coalesce(proj.construction_status, 'TBD') -- map to ISO queue_status categories
```
bendnorman, line 277: I think we should just fill the missing `construction_status` values with TBD in the airtable sheet.

src/dbcp/data_mart/projects.py
```diff
@@ -253,25 +245,19 @@ def _get_proprietary_proposed_offshore(engine: sa.engine.Engine) -> pd.DataFrame
...
-    proj_county_assoc as (
-        SELECT
-            project_id,
-            county_id_fips,
-            -- some counties have multiple cable landings from the same
-            -- project (different towns). I allocate the capacity equally
-            -- over the landings
-            (count(*) * 1.0 / max(n_locations))::real as frac_locations_in_county
-        FROM cable_locs
-        group by 1,2
```
bendnorman, line 274: Is this the groupby step your PR description is referring to? So is this allocation over multiple counties not happening any more or is it done in a different place now?

bendnorman (changes requested):
Thanks for making these changes! Looks like the CI failed because of an unexpected queue status in the offshore wind.

src/dbcp/data_mart/projects.py
TrentonBush: Oh ya that's a better place for it, thanks. I'll make `queue_status` with a mapping dict and leave `construction_status` as is

src/dbcp/data_mart/projects.py
TrentonBush: We could do that, but then we have to fix new/changed values in Airtable with each update. To me `.fillna` seems easier to automate and never have to think about again

src/dbcp/data_mart/projects.py
TrentonBush: Ya. The previous version simplified things by combining multiple towns in the same county (only 12 such cases) to a single record. For example, if a raw project looked like:

| project_id | raw_county | geocoded_county |
| --- | --- | --- |
| 1 | County 1 | County 1 |
| 1 | Town 1 | County 2 |
| 1 | Town 2 | County 2 |

The old output would look like:

| project_id | geocoded_county | frac_locations_in_county |
| --- |  --- | --- |
| 1 | County 1 | 0.333 |
| 1 | County 2 | 0.667 |

The new output would be:

| project_id | geocoded_county | frac_locations_in_county |
| --- |  --- | --- |
| 1 | County 1 | 0.333 |
| 1 | County 2 | 0.333 |
| 1 | County 2 | 0.333 |

LBNL was already handled in the same way as this new method (for its 8 multi-county cases), but with a drop_duplicates after, which only took the first record. So everything was inconsistent.

The downstream impact is small, mostly because this is a rare thing (only 8 in LBNL and 12 in offshore). The county aggregate MW figures won't change, but project counts will increase for the handful of affected counties.

bendnorman: Got it thanks!

src/dbcp/data_mart/projects.py
bendnorman: Sounds good 👍

bendnorman (approved):
Looks good. Thanks Trenton!

Issue #347, opened 2024-06-25
Integrate ACP data


TrentonBush:
Notes about ACP integration from the design doc:
* quarterly update
* project-level data is private; aggregates can be public
* update via CleanIQ website, no API (worth automating the scraping?)
* need to create a surrogate ID (and validate uniqueness)
* maybe would like a change log for the future
* add FIPS codes
* merge with 860m (860m takes priority in conflicts)
* make a new county-level aggregate table

PR #352, opened 2024-07-31
Transform ACP data

TrentonBush:
Adds transform code. There are a lot of columns but only the ~20 that don't start with `raw_` will probably be used.

Possibly Notable Things:
* I added a BIGINT dtype to the Pandas/SqlAlchemy/BigQuery mappings in `src/dbcp/helpers.py`
* I tried creating IDs by hashing parts of the row (`src/dbcp/transform/acp_projects.py:_make_surrogate_key`). I don't know if this was worth doing, but the idea was to try to have stable IDs across data updates, unlike the autoincrementing IDs in most of our other tables. Changes in the data will break the IDs, but I don't know what else we can do. `(ProjectName, PhaseName)` is unique but I'm not sure if we can rely on it in the future; there are collisions in each column individually but ACP *might* not allow collisions on the tuples. I blame ACP for not including an ID. You can see the the dozen or so collisions in cell 23 of `notebooks/49-tpb-acp_qa.ipynb` if you're so inclined


src/dbcp/metadata/private_data_warehouse.py
```diff
@@ -1,14 +1,93 @@
...
-    Column("private_data", String, primary_key=True),
+    Column("proj_id", BigInteger, primary_key=True),
+    Column("status", String),
+    Column("eia_plant_code", Integer),
+    Column("project_name", String),
+    Column("phase_name", String),
+    Column("phase_type", String),
+    Column("developers", String),
+    Column("owners", String),
+    Column("iso_rtos", String),
```
bendnorman, line 28: Do the values of this field align with lbnl/gridstatus `iso_region` fields? If so maybe we use the same field name?

src/dbcp/extract/fips_tables.py
```diff
@@ -9,8 +9,13 @@
...
 import dbcp
 
+# originally from https://www2.census.gov/geo/tiger/TIGER2021/
+CENSUS_URI = "gs://dgm-archive/census/tl_2021_us_county.zip"
+TRIBAL_LANDS_URI = "gs://dgm-archive/census/tl_2021_us_aiannh.zip"
 
-def _extract_census_counties(census_uri: Path) -> pd.DataFrame:
+
+@lru_cache(maxsize=1)  # county boundaries are also used in some transform modules
+def _extract_census_counties(census_uri: str) -> pd.DataFrame:
```
bendnorman, line 18: Why did you decide to add this cache decorator?

src/dbcp/transform/acp_projects.py
```diff
@@ -0,0 +1,458 @@
...
+    return out
+
+
+def _make_surrogate_key(raw_df: pd.DataFrame) -> pd.Series:
+    """Create a surrogate key from several data columns."""
+    # The surrogate key is a hash of several columns that should uniquely identify a row
+    # The main design decisions are whether to use raw data or transformed data as
+    # inputs, and how many columns to include in the hash.
+    # The transformed data risks changes in the transform functions that break backwards
+    # compatibility. The raw data risks changes in the raw data that break backwards
```
bendnorman, line 359: I think hashing some columns to make a surrogate key is a good idea! We should probably do this for the lbnl data too so we have steady ids throughout the our warehouse and data marts.

src/dbcp/transform/acp_projects.py
```diff
@@ -0,0 +1,458 @@
...
+import pandas as pd
+
+from dbcp.extract.fips_tables import CENSUS_URI, _extract_census_counties
+from dbcp.transform.helpers import add_county_fips_with_backup_geocoding
+
+
+def _rename_columns(raw_cols: pd.Index) -> pd.Index:
+    """Rename columns to snake case."""
+
+    def _camel_to_snake(name: str) -> str:
```
bendnorman, line 18: Maybe add this to a helper module?

src/dbcp/transform/acp_projects.py
```diff
@@ -0,0 +1,458 @@
...
+        "raw_iso_rtos": _col_transform_iso_rtos,
+        "raw_owner_types": _col_transform_owner_types,
+        "raw_mw_total_capacity": partial(
+            _col_transform_mw_total_capacity, full_df=trans
+        ),
+    }
+    for raw_col, transform in col_transforms.items():
+        new_col_name = raw_col[4:]  # remove 'raw_'
+        trans[new_col_name] = transform(trans[raw_col])
+    county_shapes = _extract_census_counties(CENSUS_URI).set_geometry("geometry")
```
bendnorman, line 442: It'd be nice to use dagster here so this acp asset could depend on the raw census asset.

src/dbcp/metadata/data_mart.py
```diff
@@ -725,3 +725,18 @@
     Column("description", String),
     schema=schema,
 )
+
+county_concrete_mw = Table(
+    "county_concrete_mw",
```
bendnorman, line 730: Is there a better name for this table? Concrete meaning highly likely mw?

src/dbcp/data_mart/county_concrete_mw.py
```diff
@@ -0,0 +1,111 @@
...
+from typing import Optional
+
+import pandas as pd
+import sqlalchemy as sa
+
+from dbcp.helpers import get_sql_engine
+
+
+def _get_concrete_aggs(engine: sa.engine.Engine) -> pd.DataFrame:
+    query = """
```
bendnorman, line 11: We need to decide if we want to use SQL or python for data mart transformations. I prefer using Python because that is what our team is more comfortable with and we don't get any SQL formatting or syntax highlighting with our current method. Also, if we use pandas for transformations we can more easily switch between different storage backends.

bendnorman (approved):
Looks good! I left a few style comments and questions.

src/dbcp/extract/fips_tables.py
TrentonBush: I needed to load the census county geometries a second time in order to QA the lat/lon fields in the ACP data. The geometries take like 10 seconds to load, so I thought I'd cache it so we save that huuuuge amount of time

src/dbcp/transform/acp_projects.py
TrentonBush: Yes it would!

src/dbcp/metadata/data_mart.py
TrentonBush: that name came from Yujia 🤷🏼I agree though we already have two inscrutable names ("actionable" and "nearly certain") in this realm and adding a third doesn't clarify things to me

src/dbcp/metadata/private_data_warehouse.py
TrentonBush: ah ya good catch. I'll standardize some other column names too, like `plant_id_eia` and `capacity_mw`

src/dbcp/transform/acp_projects.py
TrentonBush: I think the `_camel_to_snake` function is not quite generalized enough to to use elsewhere. It currently doesn't handle the (common) case of multiple consecutive capital letters, such as from abbreviations eg. `EIAPlantCode` (should output `eia_plant_code` but currently outputs `eiaplant_code`).

src/dbcp/data_mart/county_concrete_mw.py
TrentonBush: You're right, now I remember we had [a similar conversation](https://github.com/deployment-gap-model-education-fund/deployment-gap-model/pull/332#discussion_r1593156405) a couple months ago. I'll push this as is for the sake of time but I'll convert it to pandas next sprint

jdangerx:
@TrentonBush does this close #347 ?

PR #368, opened 2024-09-27
Ballot ready update 2024-08-28

bendnorman:
Update ballot ready data to latest CSV.


bendnorman:
This was handled in #372

PR #378, opened 2024-11-13
Remove pudl dependency

bendnorman:
This PR removes PUDL as a dependency because we pull PUDL data directly from AWS and only use the python package to add fips codes.


docker-compose.yaml
```diff
@@ -24,3 +24,8 @@ services:
       - .env
     ports:
       - ${POSTGRES_PORT}:5432
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U youruser"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
```
bendnorman, line 31: I added this health check because the etl in the `app` service was trying to connected to the database before it was ready.

requirements.txt
```diff
@@ -8,11 +8,18 @@ googlemaps~=4.5.3
...
 gridstatus~=0.20.0
 s3fs>=2022.11.0
 click>=8.1.7
+numpy<2
+sqlalchemy<2
+coloredlogs~=15.0.1
+scipy~=1.14.1
+google-cloud-storage~=2.18.2
+geopandas~=1.0.1
+gcsfs~=2024.10.0
```
bendnorman, line 25: Some of these dependencies were coming from PUDL so I made them direct dependencies.

PR #383, opened 2024-11-21
Resolve BigQuery datetime issue

bendnorman:
Datetime columns were being loaded to BigQuery as integers. I think this is because the pandas `datetime[ns]` columns were being saved to the parquet files as a parquet nanosecond `Timestamp` type. BQ does not [list a conversion](https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-parquet#type_conversions) type for pyarrow nanosecond Timestamps. I guess it just converts it to an integer because that's how parquet stores timestamps.

I resolved this issue by creating pyarrow schemas from the sqlalchemy types and converting all datetime columns to use milliseconds instead of nanoseconds. It might be worth going through all these columns to figure out which ones actually need a time component.

### DIFF
--- a/src/dbcp/cli.py
+++ b/src/dbcp/cli.py
@@ -1,4 +1,5 @@
 """A Command line interface for the down ballot project."""
+
 import argparse
 import logging
 import sys
@@ -66,9 +67,9 @@ def main():
         SPATIAL_CACHE.clear()
 
     if args.etl:
-        dbcp.etl.etl(args)
+        dbcp.etl.etl()
     if args.data_mart:
-        dbcp.data_mart.create_data_marts(args)
+        dbcp.data_mart.create_data_marts()
 
 
 if __name__ == "__main__":
--- a/src/dbcp/data_mart/__init__.py
+++ b/src/dbcp/data_mart/__init__.py
@@ -5,6 +5,8 @@ import logging
 import pkgutil
 
 import pandas as pd
+import pyarrow as pa
+import pyarrow.parquet as pq
 
 import dbcp
 from dbcp.constants import OUTPUT_DIR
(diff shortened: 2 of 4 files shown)


src/dbcp/cli.py
```diff
@@ -66,9 +67,9 @@ def main():
         SPATIAL_CACHE.clear()
 
     if args.etl:
-        dbcp.etl.etl(args)
+        dbcp.etl.etl()
     if args.data_mart:
-        dbcp.data_mart.create_data_marts(args)
```
bendnorman, line 71: These args weren't being used. Also, these cli commands are reworked in #366.

src/dbcp/data_mart/__init__.py
```diff
@@ -64,8 +66,8 @@ def create_data_marts(args):  # noqa: max-complexity=11
     with engine.connect() as con:
         for table in metadata.sorted_tables:
             logger.info(f"Load {table.name} to postgres.")
-            df = enforce_dtypes(data_marts[table.name], table.name, "data_mart")
-            df = dbcp.helpers.trim_columns_length(df)
+            df = dbcp.helpers.trim_columns_length(data_marts[table.name])
+            df = enforce_dtypes(df, table.name, "data_mart")
```
bendnorman, line 70: We should trim the column names before we apply dtypes because we pull the type information from the sqlalchemy tables which use the postgres column names.

src/dbcp/helpers.py
```diff
@@ -90,12 +118,16 @@ def enforce_dtypes(df: pd.DataFrame, table_name: str, schema: str):
...
-        col.name: SA_TO_PD_TYPES[str(col.type)]
-        for col in table.columns
-        if col.name in df.columns
-    }
-    return df.astype(dtypes)
+    for col in table.columns:
+        # Add the column if it doesn't exist
+        if col.name not in df.columns:
+            df[col.name] = None
+        df[col.name] = df[col.name].astype(SA_TO_PD_TYPES[str(col.type)])
```
bendnorman, line 125: Pyarrow throws an error if a column exists in the schema but not in the dataframe.

PR #385, opened 2024-11-22
Update ACP and EIA 860m 2024 Q3

bendnorman:
This PR
- updates the ACP and EIA 860m pipelines to use data from 2024 Q3
- adds a couple of notebooks for comparing old and new eia860m and acp data
- Adds a `pudl_eia860m_status_codes` table that has descriptions of the operational status codes in the eia860m table
- Uses `s3://pudl.catalyst.coop` instead of `gs://parquet.catalyst.coop`

Because we're doing a bunch of data updates right now, I'm going to create a separate PR that sets up a proper documentation framework and write update instructions.


src/dbcp/metadata/data_mart.py
```diff
@@ -46,6 +47,7 @@
     Column("infra_total_proposed_pm2_5_tonnes_per_year", Float),
     Column("battery_storage_existing_capacity_mw", Float),
     Column("battery_storage_existing_facility_count", Integer),
+    Column("battery_storage_existing_co2e_tonnes_per_year", Float),
```
e-belfer, line 50: Can you explain where this new column is coming from? Presuming it's PUDL, but seems like it should get added to the data mart documentation?

e-belfer (approved):
One request to update docs on the new column you added to the data warehouse, but otherwise looks good, ran as expected and all the data appears as expected.

Did you look into the few ~200% capacity percentage changes at the ISO level for different statuses? I'm presuming this is due to small capacity values in previous time periods but would just be good to confirm this isn't unexpected.

src/dbcp/metadata/data_mart.py
bendnorman: There is only one battery plant that has co2 emissions associated I think because it is colocated with a fossil fuel plant (plant_id_eia = 6377). TBH I'm not sure why it's popping up now given this battery and fossil generator have [been in the data for a while](https://data.catalyst.coop/pudl/core_eia923__monthly_generation_fuel?plant_id_eia__exact=6377). Maybe EIA assigned gave them the same plant id in this latest release? I would also expect there to be multiple battery + fossil generator plants. 

Given this is only one plant I don't think it's worth diving super deep. I'll add some docs!

src/dbcp/constants.py
```diff
@@ -103,7 +103,7 @@
 
 FIPS_CODE_VINTAGE = 2020
 
-PUDL_LATEST_YEAR = 2022
+PUDL_LATEST_YEAR = 2023
```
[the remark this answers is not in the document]
bendnorman: I'll add a docstring!

bendnorman:
I think the large capacity changes in the ISO are due to large projects moving along the ISO queue process. I calculated the pct change in number of projects and all the status codes are within 30% which makes more sense.

PR #394, opened 2025-01-22
Use geocodio instead of google maps

bendnorman:
This PR replaces the Google Maps geocoder with the Geocodio API. Geocodio is cheaper and allows us to cache API calls.

## Validation checks
- 97% of counties in `counties_wide_format` have the same `renewable_and_battery_proposed_capacity_mw` when geocoded with Google and Geocodio.
- Here are some additional metrics for ETLs that require geocoding. This table compares # of locations that have the same fips code after being geocoded by Google and Geocodio:

| dataset                        | % of locations that don’t have matching fips codes after addfips + geocoding | % of locations that needed geocoded that don’t have matching fips codes |
|----------------------------|--------------------------------------------------------------|--------------------------------------------------------------|
| offshore_wind              | 10%                                                          | 10%                                                          |
| gridstatus                 | 2%                                                           | 24%                                                          |
| eip_infrastructure         | 1%                                                           | 15%                                                          |
| columbia_local_opp         | 4%                                                           | 10%                                                          |
| nrel_wind_solar_ordinances | 1%                                                           | 18%                                                          |
| lbnl_iso_queue             | 1%                                                           | 36%                                                          |

These all feel like acceptable differences to me. Typically, a small percentage of locations need to be geocoded because `addFips` does most of the work. The offshore wind locations might be worth looking into but we're about to update this dataset so I figured we could do it then.

## Questions
- Do you think these validation checks are reasonable? Anything you think I should dig into a little more?
- Do you think it's worth keeping the Google Maps API logic around?


src/dbcp/transform/geocodio.py
```diff
@@ -0,0 +1,127 @@
...
+            if locality_type == "place":
+                locality_name = ad.address_components.city
+                locality_type = "city"
+            elif locality_type == "county":
+                locality_name = ad.address_components.county
+            else:
+                locality_name = ""
+            results_df.append(
+                [locality_name, locality_type, ad.address_components.county]
+            )
```
bendnorman, line 86: Might be cleaner to add this logic to the pydantic classes.

src/dbcp/transform/google_maps.py
```diff
@@ -202,3 +214,61 @@ def _get_geocode_response(
         return response[0]
     except IndexError:  # empty list = not found
         return {}
+
+
+def _geocode_row(
```
bendnorman, line 219: I moved this logic from `dbcp.transform.helpers` because it is specific to the Google Maps API.

src/dbcp/transform/google_maps.py
```diff
@@ -202,3 +214,61 @@ def _get_geocode_response(
...
+        client=geocoder,
+        state_col=state_col,
+        locality_col=locality_col,
+    )
+    new_cols.columns = [
+        "geocoded_locality_name",
+        "geocoded_locality_type",
+        "geocoded_containing_county",
+    ]
+    return new_cols
```
bendnorman, line 274: I moved this logic from `dbcp.transform.helpers` because it is specific to the Google Maps API.

src/dbcp/transform/helpers.py
```diff
@@ -241,66 +257,66 @@ def parse_dates(series: pd.Series, expected_mean_year=2020) -> pd.Series:
         return multiformat_string_date_parser(series)
 
 
-def _geocode_row(
-    ser: pd.Series, client: GoogleGeocoder, state_col="state", locality_col="county"
-) -> List[str]:
-    """Function to pass into pandas df.apply() to geocode state/locality pairs.
+def _geocode_and_add_fips(
```
bendnorman, line 260: I moved some of the geocoding logic around so I could compare the Google and Geocodio results.

src/dbcp/transform/helpers.py
```diff
@@ -332,7 +356,8 @@ def add_county_fips_with_backup_geocoding(
         with_fips["geocoded_locality_name"] = with_fips[locality_col]
         with_fips["geocoded_locality_type"] = "county"
         with_fips["geocoded_containing_county"] = with_fips[locality_col]
-        return with_fips
+        # attach to original df
```
bendnorman, line 359: This was a bug. We need to add the original df back to the fipsified dataframe.

src/dbcp/transform/helpers.py
```diff
@@ -343,45 +368,40 @@ def add_county_fips_with_backup_geocoding(
...
+        logger.info("---------------------")
+        logger.info(
+            f"---- pct of geocoded fip failures that don't match: {(~county_eq).sum() / len(comp)}"
+        )
+        logger.info(
+            f"---- pct of all records that don't have the same county: {(~county_eq).sum() / len(state_locality_df)}"
+        )
+        logger.info("---------------------")
+
+    filled_fips = geocodio_df
```
bendnorman, line 402: This is logic to compare Google and Geocodio. I figured we could remove the Google logic once the changes have settled.

src/dbcp/transform/helpers.py
```diff
@@ -241,66 +257,66 @@ def parse_dates(series: pd.Series, expected_mean_year=2020) -> pd.Series:
...
+        state_col=state_col,
+        locality_col=locality_col,
+    )
 
-    Args:
-        state_locality_df (pd.DataFrame): dataframe with state and locality columns
-        state_col (str, optional): name of the column of state names. Defaults to 'state'.
-        locality_col (str, optional): name of the column of locality names. Defaults to 'county'.
+    # recombine deduped geocoded data with original nan_fips
+    geocoded_deduped_nan_fips = pd.concat(
```
jdangerx, line 294: Is the reason we put the empty rows in while geocoding so that this concat can still work?

test/unit/test_geocoding.py
jdangerx, line 1: Should we also test `add_county_fips_with_backup_geocoding` to make sure that we're combining the addfips results with geocoding results properly?

jdangerx (changes requested):
Some non-blocking suggestions about tightening up the type definitions etc.

It would maybe be nice to add some tests of the higher-level functionality beyond just "does the API return what we expect?" (which is also a valuable test!)

Re: the validation tests - is this the right interpretation of the results?

* Out of all N locations in the gridstatus ETL, there were M locations that didn't have a FIPS  code from `addfips`. Of those M, 24% of those had a different FIPS code when using Geocodio vs. Google. 2% of the N total locations had a mismatched FIPS code. Which also means that about M is about 1/12 of N.

* Also, for the offshore wind data, that means that none of the original locations got their FIPS code from `addfips` - and 10% of the geocoded results are different between Geocodio and Google.

If that's all true, the diff does seem kind of high to me, but also causes a pretty small absolute change in codes. I would like to see, for instances where Google and Geocodio disagree, what the actual disagreement is - are they geocoding to counties that are right next to each other? Where is the actual lat/long, is it close to a county edge?

src/dbcp/transform/geocodio.py
```diff
@@ -0,0 +1,133 @@
+"""Geocodio geocoding functions."""
+
+import os
+from pathlib import Path
+
+import pandas as pd
+from geocodio import GeocodioClient
+from geocodio.exceptions import GeocodioAuthError
+from joblib import Memory
```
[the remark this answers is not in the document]
bendnorman: haha agreed!

src/dbcp/transform/geocodio.py
```diff
@@ -0,0 +1,133 @@
...
+from pydantic import BaseModel
+
+geocoder_local_cache = Path("/app/data/geocodio_cache")
+# create geocoder_local_cache if it doesn't exist
+geocoder_local_cache.mkdir(parents=True, exist_ok=True)
+assert geocoder_local_cache.exists()
+# cache needs to be accessed outside this module to call .clear()
+# limit cache size to 100 KB, keeps most recently accessed first
+GEOCODER_CACHE = Memory(location=geocoder_local_cache, bytes_limit=2**19)
```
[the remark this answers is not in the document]
bendnorman: Ah thanks for the catch. The comment and `bytes_limit` must have become out of sync a while ago. I updated both comments in the `geocodio` and `google_maps` modules.

src/dbcp/transform/geocodio.py
```diff
@@ -0,0 +1,133 @@
...
+import os
+from pathlib import Path
+
+import pandas as pd
+from geocodio import GeocodioClient
+from geocodio.exceptions import GeocodioAuthError
+from joblib import Memory
+from pydantic import BaseModel
+
+geocoder_local_cache = Path("/app/data/geocodio_cache")
```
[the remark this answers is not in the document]
bendnorman: True! We've been hard coding these paths because this ETL should only be run in a docker container which has a consistent file structure. I think it'd be wise to use an env var anyway! There are lots of locations in the code that reference `/app/data` so I'll make this change in a separate PR.

src/dbcp/transform/geocodio.py
```diff
@@ -0,0 +1,133 @@
...
+assert geocoder_local_cache.exists()
+# cache needs to be accessed outside this module to call .clear()
+# limit cache size to 100 KB, keeps most recently accessed first
+GEOCODER_CACHE = Memory(location=geocoder_local_cache, bytes_limit=2**19)
+
+
+class AddressComponents(BaseModel):
+    """Address components from Geocodio."""
+
+    number: str = ""
```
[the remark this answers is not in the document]
bendnorman: I don't think [the docs](https://www.geocod.io/docs/#address-components) say anything about values that will always be present.

src/dbcp/transform/geocodio.py
```diff
@@ -0,0 +1,133 @@
...
+    county: str = ""
+    state: str = ""
+    zip: str = ""  # noqa: A003
+    country: str = ""
+
+
+class Location(BaseModel):
+    """Location from Geocodio."""
+
+    lat: float = 0.0
```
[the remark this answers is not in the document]
bendnorman: Good point! Removed the default value for these attributes.

src/dbcp/transform/geocodio.py
```diff
@@ -0,0 +1,133 @@
...
+        results = client.geocode(batch["address"].tolist())
+    except GeocodioAuthError:
+        raise GeocodioAuthError(
+            "Geocodio API key is invalid or you hit the daily geocoding limit which you can change in the Geocodio billing tab."
+        )
+
+    results_df = []
+    for result in results:
+        if "error" in result:
+            results_df.append(["", "", ""])
```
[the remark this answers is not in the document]
bendnorman: This is handling the case when the API can't geocode an address. I updated it to be a list of `None`'s.

src/dbcp/transform/helpers.py
bendnorman: Yes it's so we can preserve the index. Do you think it'd be cleaner to not add empty rows?

jdangerx: Yeah, I think so, and intuitively seems like it would make the joining a little simpler. But it's not a blocking concern I think - refactoring would require (a) adding tests for this behavior and then (b) figuring out what sequence of joins etc. works for this use case, so it's not like it's a 5-minute dealio. Might be a good follow-up PR though.

bendnorman:
You understanding of the metrics is correct. I can do some more digging as to why google and geocodio produce different results but I couldn't find any consistent patterns.

What were you thinking for additional tests? Here are some ideas:
- We're catching geocodio authentication errors properly
- The geocodio pydantic class types are working as expected
- You suggestion: test we're combining the addfips and geocoded results correctly

jdangerx:
OK, if you've already done that digging I'm good to merge - we should probably mention that to DBCP to see if they want us to dig more or not.

As for more tests, I do think that testing `add_county_fips_with_backup_geocoding` is a good thing to do since it pulls a bunch of different functionality together and is a commonly used internal library function. I'm good with that being in this PR or as a follow-up PR.

jdangerx (approved):

bendnorman:
Good news! In the previous validation analysis, I compared county names instead of fips codes. Turns out Geocodio and Google are geocoding to the correct entity but they use different spellings. For example, in the GS and LBNL data there were about a hundred locations in Virginia that geocode to [independent cities](https://en.wikipedia.org/wiki/List_of_cities_and_counties_in_Virginia). Google geocodes "City of Chesapeake" to "Chesapeake City" and Geocodio geocodes it to "Chesapeake". Add fips returns the same fips code for both of these. The new results are much more acceptable.

| dataset                    | % of locations that don’t have matching fips codes after addfips + geocoding | % of locations that needed geocoded that don’t have matching fips codes |
|----------------------------|--------------------------------------------------------------|--------------------------------------------------------------|
| offshore_wind              | 3%                                                           | 3%                                                           |
| gridstatus                 | <1%                                                          | 4%                                                           |
| eip_infrastructure         | <1%                                                          | 8%                                                           |
| columbia_local_opp         | 4%                                                           | 10%                                                          |
| nrel_wind_solar_ordinances | <1%                                                          | <1%                                                          |
| lbnl_iso_queue             | <1%                                                          | 7%                                                           |

PR #395, opened 2025-01-27
Create DATA_DIR constant that is pulled from env var

bendnorman:
This PR:
- Replaces hardcoded `"/app/data"` references with a constant that pulls a path from an env var defined in `.env`
- Updates Google Maps API error message
- Removes default.env because all env vars are set locally or in `.env`


default.env
jdangerx, line 1: non-blocking: If I wanted to change `DATA_DIR` or somesuch could I put that in `local.env`?

src/dbcp/extract/local_opposition.py
```diff
@@ -91,7 +90,7 @@ def _remove_intro(
                 return paragraphs[idx:]
         raise ValueError("Could not find starting state")
 
-    def _parse_values(self, text: str) -> None:
+    def _parse_values(self, text: str) -> None:  # noqa: C901
```
jdangerx, line 93: non-blocking: it might be worth pulling each section handler out into its own function, then using `match/case` or a dictionary to handle dispatch based on each section.

Are the sections always in a specific order? Or can they be a random jumble?

default.env
bendnorman: In the .env file!

PR #400, opened 2025-02-05
Update BallotReady Q1 data

e-belfer:
Update BallotReady data to use `climate_partners_upcoming_races_with_counties_20250204.csv`.

The PR includes the following changes:
- Update the input data
- Update the notebook
- Remove duplicate specifications of the UI
- Fix a bunch of schema TODOs relating to null county_id_fips and foreign keys
- Handle duplicated county names
- Incorporate geocoding fix from #401


src/dbcp/transform/ballot_ready.py
```diff
@@ -149,9 +151,13 @@ def _explode_counties(raw_ballot_ready: pd.DataFrame) -> pd.DataFrame:
...
     ballot_ready["counties"] = (
-        ballot_ready.counties.str.replace('"', "").str[1:-1].str.split(", ")
-    )
-
+        ballot_ready.counties.str.replace('"', "")
+        .str[1:-1]
+        .str.split(",")
+        .apply(set)
+        .apply(list)
+    )  # The 2/2025 update has tons of duplicate counties in the counties column list,
```
bendnorman, line 159: Any idea why there are way more duplicate counties?

src/dbcp/transform/ballot_ready.py
```diff
@@ -214,18 +223,17 @@ def _explode_counties(raw_ballot_ready: pd.DataFrame) -> pd.DataFrame:
...
+        (ballot_ready.geo_id.str.len().isin([5, 10, 15, 16]))
         & (ballot_ready.county_id_fips.notnull())
     ]
     county_match = geo_ids.geo_id.str[0:5] == geo_ids.county_id_fips
     logger.info(
-        f"County FIPS codes:{sum(county_match)} of {len(county_match)} geocoded state FIPS IDs match the Ballot Ready data ({sum(county_match)/len(county_match):.0%})"
+        f"County FIPS codes:Of {len(county_match)} geocoded state FIPS IDs compared to the Ballot Ready data, {sum(county_match)} match ({sum(county_match)/len(county_match):.0%})"
     )
-    assert sum(county_match) / len(county_match) > 0.85
+    assert sum(county_match) / len(county_match) > 0.75
```
bendnorman, line 236: Any idea why this number went down? Is this even a problem given we aren't using geo ids rn?

src/dbcp/transform/ballot_ready.py
```diff
@@ -264,7 +277,7 @@ def transform(raw_dfs: dict[str, pd.DataFrame]) -> dict[str, pd.DataFrame]:
     from dbcp.extract.ballot_ready import extract
 
     # is this URI still up to date?
```
bendnorman, line 279: I think this comment should be removed? Also, we could make the URI a constant in `dbcp.extract.ballot_ready` so we don't have to update it in multiple places.

bendnorman (commented):
A few non-blocking questions/comments.

It would be nice to get a synopsis of the changes BR made since the last update so we have something to validate against. For example, it would be nice to know which states they added races for to compare with this map you created: 
image

bendnorman (approved):

src/dbcp/transform/ballot_ready.py
e-belfer: This isn't a problem, since geo_ids don't guarantee that they have County FIPS IDs, only state FIPS IDs. But anecdotally seems like there's way more straying from the Census block tract format to favor text inputs in the GEOID column, as opposed to the standard state + county + block ID formats.

Matches here are coarsely based on string lengths that include County FIPS IDs from the Census, so the less standardized this field is the less matches we'll see.

src/dbcp/transform/ballot_ready.py
e-belfer: I'm not sure, it might've been something about the export or a data quality problem on their end but it's widespread and odd.

PR #402, opened 2025-02-12
Pull manual ordinance data from Airtable using new infrastructure

bendnorman:


src/dbcp/commands/publish.py
```diff
@@ -80,39 +80,40 @@ def load_parquet_files_to_bigquery(
...
-            table.labels = labels
-            client.update_table(table, ["labels"])
+        # add a label to the table
+        labels = {"version": version}
+        table = client.get_table(table_ref)
+        table.labels = labels
+        client.update_table(table, ["labels"])
 
-            logger.info(f"Loaded {blob.name} to {dataset_id}.{table_name}")
+        logger.info(f"Loaded {blob.name} to {dataset_id}.{table_name}")
```
bendnorman, line 116: Some improvements Dazhong [suggested](https://github.com/deployment-gap-model-education-fund/deployment-gap-model/pull/366#discussion_r1842989054) in the initial Airtable PR.

src/dbcp/extract/manual_ordinances.py
```diff
@@ -1,30 +1,44 @@
...
+
 import pandas as pd
-from pandas_gbq import read_gbq
+
+from dbcp.extract.helpers import cache_gcs_archive_file_locally, extract_airtable_data
 
 
-def extract() -> dict[str, pd.DataFrame]:
+def extract(counties_uri: str) -> dict[str, pd.DataFrame]:
     """Extract manually maintained ordinances from BigQuery.
```
jdangerx, line 9:
```suggestion
    """Extract manually maintained ordinances from AirTable snapshot.
    
    Arguments:
        counties_uri: object name within `dgm-archive` GCS bucket with a snapshot of the AirTable table.
```

src/dbcp/extract/manual_ordinances.py
```diff
@@ -1,30 +1,44 @@
...
-    CASE
-        WHEN ordinance_status IS NULL THEN NULL
-        WHEN ordinance_status IN ('Ordinance/moratorium change in process', 'Prohibitive ordinance/moratorium') THEN TRUE
-        ELSE FALSE
-    END as ordinance_via_self_maintained
-    FROM `local-jobs-econ-dev-fund.airtable_data.county_permitting_info`
-    """,
-        project_id=project_id,
-        credentials=credentials,
+    counties_uri, counties_generation_num = str(counties_uri).split("#")
```
jdangerx, line 15: nit: It seems like it might be nicer to have the URI and generation number be separate params, then we don't have to have this `#` split situation.

src/dbcp/etl.py
```diff
@@ -186,7 +186,16 @@ def etl_gridstatus_isoqueues():
 
 def etl_manual_ordinances() -> dict[str, pd.DataFrame]:
     """ETL manually maintained ordinances."""
-    raw_dfs = dbcp.extract.manual_ordinances.extract()
+    county_level_control_uri = (
+        "airtable/Clean Build State Permitting Analysis/County Level Control.json"
+    )
+
+    es = ExtractionSettings.from_yaml("/app/dbcp/settings.yaml")
+    es.update_archive_generation_numbers()
```
jdangerx, line 194: For my understanding: this function does:

* if there wasn't a generation specified in settings.yaml, finds the latest generation num so everything downstream of this happens with a concrete generation number
* otherwise, grab the pinned one

Is that right?

If so, we might also consider calling this `update_archive_generation_numbers` *in* `from_yaml` so that we never forget to do it.

src/dbcp/settings.yaml
```diff
@@ -10,3 +10,9 @@
     table_id: tblVU9JPtahIwLbny
   name: airtable/Offshore Wind Locations Synapse Version/Projects.json
   pinned: false
+- generation_num: 1739398209345752
```
jdangerx, line 13: FMI: when/how do we decide to update these pinned versions?

jdangerx (approved):
I trust that this will work. Have some documentation tweaks + just some questions for my own edification, but I'm OK shipping this without another review :)

bendnorman:
Thanks for the review! I added some new thoughts about this PR [to this issue](https://github.com/catalyst-cooperative/tech-private/issues/35#issuecomment-2731198633).

PR #418, opened 2025-03-25
Offshore wind archive fix

katie-lamb:
The `update-data` action (run on `main`) didn’t successfully pull in the latest offshore wind info from Airtable. Specifically ‘Coastal Virginia Offshore Wind’, ‘Revolution Wind (RI)’, ‘New England Wind Phase 2’ seem to have mismatch between BQ and Airtable. For example, the "Overall project status" for Coastal Virginia Offshore Wind is Late Stage in Airtable and Under Way in the GCS archive + BQ data mart tables.

This is what I found:

* I think the Airtable base with offshore wind data updates is called "Offshore Wind Locations DBCP version". It seems like the archiver in main (and dev) is pointing toward a different Airtable base (maybe the old Synapse version one?)
* If this is the case, and the DBCP version Airtable base is what we're supposed to be making an archive of / pulling from, then I updated the `base_id` variable to be the ID for this new Airtable base (found from the URL of the base). 
* I had to create a new Airtable personal access token that has access to this Airtable base in order to successfully run `make archive_all`.
* I was able to create a `test` archive [in GCS](https://console.cloud.google.com/storage/browser/dgm-archive/test/TEST:%20Offshore%20Wind%20Locations%20Synapse%20Version?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&authuser=2&project=dbcp-dev-350818) which looks correct (I'm looking at Projects.json)
* I tried running the `update-data` action on this branch but am getting an Airtable permissions error - it seems like the Github runners access token doesn't have access to this Airtable base.

To Do:

- [x] Update the Airtable API Key for the Github runner to a key which has access to this new Airtable base. This should be done by someone with access to all the various Airtable bases (I only have access to this Offshore wind one at this point). Alternatively, we could make a new token/Github secret with specific access to this offshore wind base so that we don't need to change it for the other bases.
- [x] Verify that the "Offshore Wind Locations DBCP version" Airtable base is indeed what we should be archiving / pulling data from

Questions:
* Should the base branch for this be `dev` or `main`?


src/dbcp/settings.yaml
```diff
@@ -1,12 +1,12 @@
...
-    schema_generation_number: '1741396058325017'
+    schema_generation_number: '1742928586963059'
     table_id: tblu7b4Yj58Iq2xCF
   name: airtable/Offshore Wind Locations Synapse Version/Locations.json
   pinned: false
-- generation_num: 1741396058788973
+- generation_num: 1742928592428151
   metadata:
-    schema_generation_number: '1741396058325017'
+    schema_generation_number: '1742928586963059'
```
katie-lamb, line 9: hmmm im not sure if these should be updated or not? I think they should? I'm changing which base the archive actually points at, but I think the system of updating the generation number stands.

src/dbcp/settings.yaml
```diff
@@ -1,12 +1,12 @@
-- generation_num: 1741396059395591
+- generation_num: 1742928593170421
   metadata:
-    schema_generation_number: '1741396058325017'
+    schema_generation_number: '1742928586963059'
     table_id: tblu7b4Yj58Iq2xCF
   name: airtable/Offshore Wind Locations Synapse Version/Locations.json
```
katie-lamb, line 5: Maybe these names should be updated to Offshore Wind Locations DBCP Version? Maybe it happens automatically

jdangerx:
Base branch should be dev. I'll talk to Joey about the Airtable permissions.

PR #419, opened 2025-04-02
Update `update-data` workflow to run weekly

katie-lamb:
I updated the `update-data` workflow to run weekly at midnight on Sunday. Do we want to change what day/time it's run?


katie-lamb:
@jdangerx verified that expenses from weekly runs are okay

marianneke (approved):

PR #421, opened 2025-05-23
Add helper branch data comparison

marianneke:
Comparing the data outputs between git branches is a bit manual and error prone. This PR adds a script to automate parts of this process in order to make it a lot easier to obtain the data to make branch comparisons of a target branch to a base branch (`dev` by default).

In short, it checks out the target branch, and then the base branch, and for each branch:
- Runs make all to generate the data output files
- Copies the parquet files that are generated during the above run to compare to a newly created temporary data folder with the branch name as subfolder

This means we can create a relatively standardized notebook for reading the data in these folders to make comparisons

TODO:
- [x] Add unit tests

Out of scope:
- think about cleanup
- standardize some comparison functions

Notes:
If you want to run the tests, you need to reinstall the dbcp package using `pip install -e .` to make the scripts available. The googlemaps package may cause some issues with this, so in that case you may need to install googlemaps separately first.


marianneke:
> I see that there is already a CLI in this repo with a few commands (see `cli.py` and `commands/archive.py`). This might belong as a CLI command instead of as a script. Thoughts on moving it there?

My thoughts were that the CLI is for the data processing itself, and this is more of a convenience tool for developing, so the two should not necessarily be mixed. What do you think?

PR #432, opened 2025-07-04
Update dependencies and `pyproject.toml`

katie-lamb:
In the process of running a data update I thought it would be helpful to update some dependencies and simplify where we name dependencies. Initially it was spread between `requirements.txt` and `setup.py`. But `setup.py` was recently removed. I thought it would be simpler if everything was in `pyproject.toml`. This PR makes the following changes:

- Removes PUDL as a dependency since we're now just accessing PUDL tables directly from AWS
- Updates some key dependencies to their latest version - namely python and pandas
- Removes `requirements.txt` in favor of putting all dependencies under `pyproject.toml`. It seems like the main advantage of `requirements.txt` was having one place where Docker could install all development dependencies, but I instead had Docker just download the `dbcp` package + all the dependencies in `pyproject.toml`. I'm not that familiar with Docker so I'm not sure if this is problematic or not the standard convention.

Remaining To Do:

- [ ] Add more to the DBCP documentation google doc about comparison notebooks and running them from within the docker container
- [ ] Check on things like dependabot to see if it's keeping dependencies updated
- [ ] Separate out development dependencies from package dependencies


katie-lamb:
Handled in #501

PR #433, opened 2025-07-07
ACP data update

katie-lamb:
ETL updated ACP data with a new quarter of data. Changes made in this PR:

* Handles the "Investor Owned" owner type value by replacing it with "IPP"
* Adds a check for the expected number of "Investor Owned" owner type values so that in the future we can catch if this should actually map to something like "Utility: IOU" instead of IPP


marianneke (approved):
lgtm! Will be interesting to see if the same things can be found in the 860m data

PR #435, opened 2025-07-12
Merge dev into main

marianneke:
Add latest changes to production.

In the future, we will work from the main branch directly and label release branches as in PUDL, but until then we need to merge dev into main in order to release changes in production


src/scripts/branch_compare_helper.py
```diff
@@ -85,19 +85,33 @@ def main(target_branch, base_branch="dev"):
...
+
+        target_branch = args.target_branch or get_current_branch()
+        base_branch = args.base_branch
+    compare_branches(target_branch=target_branch, base_branch=base_branch)
 
-    target = args.target_branch or get_current_branch()
-    main(target_branch=target, base_branch=args.base_branch)
+
+# make it possible to call the script from the CLI entry
+# point without args.
```
jdangerx, line 110: drive-by: I've always chucked the `argparse.ArgumentParser()` definition inside this block instead of inside `main()` since we only have to parse args when we're running the script directly

jdangerx (approved):

PR #437, opened 2025-07-16
Create changelog of ACP data

### LINKED ISSUE #436: Create ACP Changelog Table
Similar to the EIA 860 data, we want to create a changelog that records changes in ACP projects as the data refreshes. Whenever any of the reported ACP project data is changed for a record, there will be a new changelog record with a new `report_date` that gives the quarter that the data was last recorded. All the attributes for the record will be merged onto the changelog.

### To do list:

- [x] Investigate creation of changelog table in PUDL repo and see what code can be reused
- [ ] Make a new data warehouse table that records changes with the ACP table
  - [x] Read in all old data from GCS archives in extract module
  - [x] Add in a column that tracks time, like a `report_date`
  - [x] Perform transformations on the raw data in the transform module
  - [ ] Experiment with using different subsets of columns to deduplicate on
  - [ ] Deduplicate so that the first occurrence of each unique record is kept
  - [ ] Add metadata for a new data warehouse table in `dbcp.metadata.data_warehouse`
- [ ] Include documentation for ACP changelog table 
- [ ] Include documentation for EIA 860 changelog table that refers back to PUDL doc pages
- [ ] Report back to Joey with instructions for using the table

### Questions:

* Should we create the changelog from scratch every time (reading in all files from the archive and creating from raw data) or should we read in the latest version of the changelog from data warehouse tables and compare with the latest quarter of data?

katie-lamb:
Closes #436 

This PR creates a new data warehouse table `acp_projects_changelog` by adding an extract and transform function to the existing ACP modules. This preserves the `acp_projects` data warehouse table, which contains just the latest version of the data.

### Naming and Table Schema
Ideally I think we'd have three tables:

* Snapshots: raw concatenated ACP data snapshots extracted in extract module - has one date column: report_date 
* Changelog: tracks when state changes for each ID - has two date columns: valid from & valid to or e.g. report_date & valid_until_date
* Timeseries named by the frequency, e.g Yearly, Quarterly, Monthly: similar to snapshots but gives the state for each ID at certain regular intervals (whereas snapshots are reported at irregular times)

I think for now we can assume quarterly updates of ACP data, so maybe we only need a yearly and monthly timeseries table.

**Change Log**
A change in any of these columns results in a change of state and a new record in the change log table:
_proj_id, status, plant_id_eia, project_name, phase_name, resource, developers, owners, iso_region, owner_types, capacity_mw, state_id_fips, county_id_fips, geocoded_locality_name, geocoded_locality_type, geocoded_containing_county, geocoded_county_id_fips, geocoded_state_id_fips, census_county_id_fips_

Notably this doesn't include the raw columns, as meaningful columns get transformed and renamed.

### Process
* Created an extract function that reads in and concatenates all raw ACP files from the GCS archive
* Added a `report_date` column based on the date in the filename of the raw data so we know what quarter/date the record is coming from
* Passed this concatenated dataframe of snapshots into a transform function which:
  * applies the existing ACP transformation to each unique `report_date` group (i.e. each of the separate raw dataframes)
  * sorts by report date so the oldest records are first
  * drops duplicates (on a subset of columns) so that we only keep the oldest version of each unique record
* Creates a private data warehouse table with an `acp_projects_changelog` table.

### Other Changes Made
* Update naming schema of raw ACP data in bucket to `projects_{month}_{day}_{year}`? If we're assuming quarterly updates then maybe it's okay to leave naming schema as is?

### Questions / Concerns
* Should we create the changelog from scratch every time (reading in all files from the archive and creating the change log from raw data) or should we read in the latest version of a changelog from data warehouse tables and compare with the latest quarter of data to update it? I opted for the first because then we wouldn't be relying on a changelog already being present, but maybe there should be a default option to check if a changelog is present and then just compare the latest quarter to the existing changelog. 
* There's some redundancy with reading in all the different raw files for the changelog and also reading in the latest ACP data to create the existing table in a separate extract function.

### To Do:

- [x] Test creation of change log private data warehouse table
- [ ] Get rid of groupby apply with SQL query?
- [ ] Create timeseries tables - do yearly and monthly?
- [x] Decide on file name formats for the raw ACP data - update to handle `projects_{month}_{day}_{year}.csv`?
  - [x] Make a small PR that updates the extraction for the standard data mart table so that the data update workflow doesn't fail 
- [ ] fill in `NaT` in `valid_until_date`?

### DIFF
--- a/src/dbcp/etl.py
+++ b/src/dbcp/etl.py
@@ -192,7 +192,6 @@ def etl_manual_ordinances() -> dict[str, pd.DataFrame]:
 
 def etl_acp_projects() -> dict[str, pd.DataFrame]:
     """ETL ACP projects."""
-    # acp_uri = "gs://dgm-archive/acp/projects_Q1_2025.csv"
     raw_dfs = dbcp.extract.acp_projects.extract()
     transformed = dbcp.transform.acp_projects.transform(raw_dfs)
     return transformed
--- a/src/dbcp/settings.yaml
+++ b/src/dbcp/settings.yaml
@@ -1,12 +1,12 @@
-- generation_num: 1752368494432476
+- generation_num: 1754182970421151
   metadata:
-    schema_generation_number: '1752368493374644'
+    schema_generation_number: '1754182968900627'
     table_id: tbluX3aa4T2lhvRFv
   name: airtable/Offshore Wind Locations DBCP Version/Locations.json
   pinned: false
-- generation_num: 1752368493783061
+- generation_num: 1754182969430555
   metadata:
-    schema_generation_number: '1752368493374644'
+    schema_generation_number: '1754182968900627'
     table_id: tblVK1P1eYblnevqo
   name: airtable/Offshore Wind Locations DBCP Version/Projects.json
   pinned: false


src/dbcp/extract/acp_projects.py
```diff
@@ -5,6 +5,25 @@
...
+    file_paths = dbcp.extract.helpers.cache_gcs_archive_bucket_contents_locally(
+        gcs_dir_name="acp"
+    )
+    concat_df = pd.DataFrame()
+    for path in file_paths:
+        new_df = pd.read_csv(path)
+        date = pd.to_datetime("-".join(path.parts[-1].split(".")[0].split("_")[1:]))
+        new_df["report_date"] = date
+        concat_df = pd.concat([concat_df, new_df])
+    return concat_df
```
katie-lamb, line 26: This reads in all the raw ACP data and concatenates them together in preparation for making a changelog. I could update this to check whether there's already a change log in the data warehouse, and then just extract the latest version of the data to compare against the existing change log. However, then we'd have to know the date of the last version of the data that the existing change log was created from.

src/dbcp/transform/acp_projects.py
```diff
@@ -49,9 +49,9 @@ def _col_transform_status(ser: pd.Series) -> pd.Series:
         pd.NA,
     }
     is_expected = out.isin(expected_values)
-    assert is_expected.all(), (
-        f"Unexpected status values: {out[~is_expected].value_counts()}"
-    )
+    assert (
+        is_expected.all()
+    ), f"Unexpected status values: {out[~is_expected].value_counts()}"
```
katie-lamb, line 54: I think some linting stuff got updated and the pre-commit made these changes.

src/dbcp/transform/acp_projects.py
```diff
@@ -479,4 +479,30 @@ def _str_strip(ser):
...
+def _transform_acp_changelog(raw_concat_df: pd.DataFrame) -> pd.DataFrame:
+    """Create a changelog of the ACP data over time."""
+    transformed_concat_df = raw_concat_df.groupby(
+        "report_date", group_keys=False
+    ).apply(
+        _transform_acp_projects
+    )  # TODO: hopefully this groupby apply isn't too slow
+    transformed_concat_df = transformed_concat_df.sort_values(by="report_date")
+    dedupe_cols = transformed_concat_df.columns - "report_date"
+    deduped_df = raw_concat_df.drop_duplicates(subset=dedupe_cols)
```
katie-lamb, line 494: haven't tested any of this, but this is roughly what I'm thinking

marianneke: I'm guessing SQL with some window functions is really all you need

src/dbcp/transform/acp_projects.py
```diff
@@ -479,4 +486,64 @@ def _str_strip(ser):
...
+
+
+def _transform_acp_snapshots_to_changelog(
+    raw_snapshots_df: pd.DataFrame,
+) -> pd.DataFrame:
+    """Create a changelog of the ACP data over time."""
+    trans_df = raw_snapshots_df.reset_index(drop=True).convert_dtypes()
+    trans_df = (
+        trans_df.groupby("report_date")
+        .apply(_clean_col_names_and_create_id)
```
katie-lamb, line 520: this isn't fast, but it's also not that slow. i like your suggestion to use a SQL query + window functions instead but I don't really trust my SQL knowledge enough to attempt to refactor `_clean_col_names_and_create_id` / thought I'd just be wasting time trying to figure it out. thoughts on how hard this might be?

marianneke:
Question about this:
> A change in any of these columns results in a change of state and a new record in the change log table:
proj_id, status, plant_id_eia, project_name, phase_name, resource, developers, owners, iso_region, owner_types, capacity_mw, state_id_fips, county_id_fips, geocoded_locality_name, geocoded_locality_type, geocoded_containing_county, geocoded_county_id_fips, geocoded_state_id_fips, census_county_id_fips

I would assume a changelog is the log of changes to the state (consisting in this case of several columns) of an entity. This entity is identified by one of more ID columns. I would assume in this case that `project_id` and `plant_id_eia` are columns that identify the entity, and are therefore not part of the state of that entity. Is that correct?

marianneke:
> Update naming schema of raw ACP data in bucket to projects_{month}_{day}_{year}? If we're assuming quarterly updates then maybe it's okay to leave naming schema as is?

Apologies if this is repetitive, but I definitely support making this change in naming convention, since the frequency of updates is likely to change

src/dbcp/extract/acp_projects.py
```diff
@@ -19,7 +50,7 @@ def extract(projects_path: Path) -> dict[str, pd.DataFrame]:
     """
     acp_raw_dfs = {}
     projects_path = dbcp.extract.helpers.cache_gcs_archive_file_locally(projects_path)
-
     acp_raw_dfs["raw_acp_projects"] = pd.read_csv(projects_path)
+    acp_raw_dfs["raw_acp_projects_snapshots"] = _extract_acp_projects_snapshots()
```
marianneke, line 54: I think this can replace the "raw_acp_projects" property, where the latest snapshot is the one being used in place of "raw_acp_projects". Do note that the table reading the latest snapshot really should have `_current` in its name according to our naming convention

src/dbcp/metadata/private_data_warehouse.py
```diff
@@ -14,6 +15,86 @@
...
+    Column("raw_tracking_types", String),
+    Column("raw_ilrs", Float),
+    Column("raw_nameplate_mws", String),
+    Column("raw_storage_energies", String),
+    Column("raw_storage_durations", String),
+    schema=schema,
+)
+
 acp_projects = Table(
     "acp_projects",
```
marianneke, line 99: So this table really should have a name that indicates what type of "time" it contains, meaning is it changelog, snapshots, current state or timeseries? I believe this is current state

src/dbcp/extract/acp_projects.py
```diff
@@ -1,10 +1,41 @@
...
+            raise ValueError(
+                f"ACP filename '{filename}' is not formatted as expected (should match 'projects_Q{{1-4}}_{{YYYY}}'). "
+                f"Unable to add a date to this ACP snapshot -- update the filename or the way that these filenames are parsed."
+            )
+
+        quarter = match.group(1)
+        month = quarter_to_month[quarter]
+        year = int(match.group(2))
+        snapshot_df["report_date"] = pd.to_datetime(f"{year}-{month:02d}-01")
+        concat_df = pd.concat([concat_df, snapshot_df])
```
marianneke, line 37: This is slow since you call `pd.concat` repeatedly. I think it's faster to just add all new dfs to a list and call `pd.concat` once at the end

src/dbcp/extract/acp_projects.py
```diff
@@ -1,10 +1,41 @@
...
+    )
+    concat_df = pd.DataFrame()
+    quarter_to_month = {"1": 1, "2": 4, "3": 7, "4": 10}
+    for path in file_paths:
+        snapshot_df = pd.read_csv(path)
+        filename = path.parts[-1].split(".")[0]
+        match = re.match(r"projects_Q([1-4])_(\d{4})", filename)
+        if not match or match.lastindex < 2:
+            raise ValueError(
+                f"ACP filename '{filename}' is not formatted as expected (should match 'projects_Q{{1-4}}_{{YYYY}}'). "
```
marianneke, line 29: If we move to monthly snapshots, should we then also accept both quaterly naming conventions and monthly, and only transform to monthly if it's the former?

src/dbcp/transform/acp_projects.py
```diff
@@ -479,4 +486,64 @@ def _str_strip(ser):
     }
     out.rename(columns=rename_dict, inplace=True)
     out.drop(columns=cols_to_drop, inplace=True)
-    return {"acp_projects": out}
+    return out
+
+
+def _clean_col_names_and_create_id(df) -> pd.DataFrame:
+    """Create surrogate key based on raw columns, then rename and assign ID.
```
marianneke, line 493: I think we should make it clear whether we are creating an ID or a state hash in order to track state changes.

src/dbcp/transform/acp_projects.py
```diff
@@ -479,4 +486,64 @@ def _str_strip(ser):
...
+        .rename(columns={"raw_report_date": "report_date"})
+    )  # TODO: hopefully this groupby apply isn't too slow
+    trans_df = _clean_columns(trans_df)
+    trans_df = trans_df.sort_values(by="report_date")
+    dedupe_cols = [
+        col
+        for col in list(trans_df.columns)
+        if (col not in ["report_date"] and not col.startswith("raw"))
+    ]
+    deduped_df = trans_df.drop_duplicates(subset=dedupe_cols)
```
marianneke, line 530: This will not work properly in the (probably unlikely) case that a state reverts to an earlier seen state. The correct way would be to add a `state_group` property just like in the SQL I showed above, which you then add to the columns the drop_duplicates command considers

marianneke (changes requested):
Your implementation would likely work in practice, but there is anedge case for which it is incorrect, which is if a state reverts to an earlier seen state. The clean way to implement is to track state changes and apply a state_group column (basically an integer that increases every time the state of the relevant columns has changed for a given project). Then you deduplicate only if this group value is the same.

More cosmetically, I'd like to include at least a change in naming of the current table for acp projects to indicate it is the current table.

src/dbcp/transform/acp_projects.py
katie-lamb: This ID creation is actually from the previous ACP table creation (not change log) code. It creates a `proj_id` surrogate key based on some of the columns, because we lack a primary key to identify each project in the raw data. 

When creating the change log, I still wanted to create + assign the `proj_id` to each row. Then, I group by this `proj_id` and hash each row (which hashes more columns than are used int he creation of the `proj_id`). So in my mind, it's necessary to have the `proj_id` (which connects projects to other ACP tables and uniquely identifies a project), and a state hash (which is a hash of all columns, not just those used to create the surrogate key `proj_id`).

src/dbcp/extract/acp_projects.py
katie-lamb: Converted to a naming format with a full date in the file name `acp_projects_{YYYY}_{MM}_{DD}`

src/dbcp/metadata/private_data_warehouse.py
katie-lamb: This is the current state table. Previously, it was the only table that we tracked with ACP data.

src/dbcp/transform/acp_projects.py
```diff
@@ -479,4 +491,75 @@ def _str_strip(ser):
...
+        trans_df[compare_cols]
+        .astype(str)
+        .apply(
+            lambda x: hashlib.sha256(str(tuple(x)).encode("utf-8")).hexdigest(), axis=1
+        )
+    )
+    trans_df["prev_hash"] = trans_df.groupby("proj_id")["row_hash"].shift()
+    out_df = trans_df[
+        (trans_df.prev_hash.isnull()) | (trans_df.row_hash != trans_df.prev_hash)
+    ]
```
katie-lamb, line 543: Changed this to hash the rows and compare to the previous row.

src/dbcp/transform/acp_projects.py
```diff
@@ -141,12 +141,19 @@ def _col_transform_owner_types(ser: pd.Series, full_df: pd.DataFrame) -> pd.Seri
...
     # not a utility company. In the future, "Investor Owned" could refer
     # to "Utility: IOU" instead of "IPP"
-    investor_owned_owner_type = out[out == "Investor Owned"]
+    n_max_investor_owned_owner_type_per_date = (
+        full_df[full_df["raw_owner_types"].str.contains("Investor Owned")]
+        .groupby("report_date")
+        .size()
+        .max()
+    )
     n_known_investor_owned = 1
```
marianneke, line 150: Not blocking, but I would be interested in how one can determine the known number of investor owned project

PR #439, opened 2025-07-28
add capacity_mw column to timeseries tables EIA

marianneke:
Adding capacity_mw as a column to the data mart timeseries tables for EIA860M.

- Leave capacity_mw column in during the query
- This means dense timeseries cannot be created using stack/unstack (which requires a Pandas Series object) but instead by explicitly creating the date spine and ID spine and creating the grid for which we want to have records, then merging on the data we have
- Add new column to docs and to metadata


katie-lamb (approved):
Looks good!

PR #443, opened 2025-08-13
Pin PUDL version to nightly

katie-lamb:
In preparation for EIA 860 monthly updates, this PR pins the PUDL version to nightly and adds a new column to the data warehouse table.

### DIFF
--- a/src/dbcp/metadata/data_warehouse.py
+++ b/src/dbcp/metadata/data_warehouse.py
@@ -591,6 +591,7 @@ pudl_generators = Table(
     Column("summer_estimated_capability_mw", Float),
     Column("supercritical_tech", String),
     Column("can_switch_oil_gas", Boolean),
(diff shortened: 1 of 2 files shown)


src/dbcp/metadata/data_warehouse.py
```diff
@@ -591,6 +591,7 @@
     Column("summer_estimated_capability_mw", Float),
     Column("supercritical_tech", String),
     Column("can_switch_oil_gas", Boolean),
+    Column("can_switch_when_operating", Boolean),
```
katie-lamb, line 594: I'm not entirely sure why switching to nightly required adding this column to the `pudl_generators` data warehouse table. It seems like this column has been present in the EIA 860 multifuel table for 6 months... might be worth more investigation if/when we have time.

marianneke (approved):
One question but not breaking, assuming builds pass

PR #462, opened 2025-10-12
Update README and update-data workflow production deployment protocol

katie-lamb:
This PR changes the way we handle deploying new data to the production tables. A couple months back, we switched from using a `dev` and `main` branch to just using `main` and updating prod tables from version tagged commits. The way we update tables in BQ is by running the `update-data` GH workflow. However, there was a disparity where pushing a tag would trigger the `update-data` workflow, but the weekly scheduled update would run on the latest released version (which could diverge from the latest pushed tag if no release was created). This PR does the following:

* Updates the README to specify that prod tables are updated from the latest version release (not the latest pushed tag)
* Updates the `update-data` workflow so that:
  * Making a new release triggers a run
  * Pushing a version tag only updates `dev` tables (not `prod` tables)
  * The weekly scheduled update runs off the latest release
  
I think ideally it would be nice to be able to manually trigger (with `workflow_dispatch`) to do a prod table update without making a new release, but setting this up was kind of confusing me, so I thought I'd start simple and get the scheduled update back up and running. Open to thoughts on improving this.

### DIFF
--- a/.github/workflows/update-data.yml
+++ b/.github/workflows/update-data.yml
@@ -148,7 +148,7 @@ jobs:
                   TARGET="prod"; REF="$TAG"
                 fi
               else
-                REF_INPUT="${{ inputs.branch_ref }}"
+                REF_INPUT="${{ github.ref_name }}"
                 TARGET="dev"; REF="${REF_INPUT:-main}"
               fi
 


.github/workflows/update-data.yml
```diff
@@ -7,6 +7,21 @@ on:
...
+      target:
+        description: 'dev or prod'
+        required: true
+        default: 'dev'
+        type: choice
+        options: [dev, prod]
+      branch_ref:
+        description: 'Branch to use for dev (ignored for prod)'
+        required: false
+        default: 'main'
```
zschira: I think this argument is probably not necessary. When you run workflows there's always the `Use workflow from` option that allows you to select a branch or tag, and then I believe you can access that with `${{ github.ref_name }}`. I've left a suggestion here, and one below that I think should allow us to remove this.

```suggestion
```

PR #484, opened 2025-11-07
Fix dependapbot failures by reworking google auth in actions

zschira:
## Overview
Our dependabot PR's are failing because they don't have access to repo secrets. We could add a dependabot secret (although I don't have access to do this myself) to mirror the Google auth credentials secret, but I decided to switch over to using workflow identity federation for authentication since it's more secure and what we are using in most of our other repos. Most of this setup I did with the `gcloud` CLI following [these instructions](https://github.com/google-github-actions/auth?tab=readme-ov-file#workload-identity-federation-through-a-service-account), then just had to modify our actions to make use of this new setup.

### DIFF
--- a/src/dbcp/commands/publish.py
+++ b/src/dbcp/commands/publish.py
@@ -139,7 +139,7 @@ class OutputMetadata(BaseModel):
     date_created: datetime = datetime.now()
 
     @validator("git_ref")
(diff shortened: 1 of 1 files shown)


katie-lamb:
This looks good! I want to test to make sure that the `update-data` action works okay, but there's a little fix I've been meaning to fix that action when running it from a branch. I'll push that fix and then can test this PR.

PR #486, opened 2025-11-18
Fix `dependabot.yml` by removing target branch

katie-lamb:
I think this fixes failing dependabot runs (logs [here](https://github.com/deployment-gap-model-education-fund/deployment-gap-model/actions/runs/19417002101/job/55547131373))

Dependabot previously ran dependency updates against the `dev` branch, but we no longer have a `dev` branch and instead just update `dev` and `prod` tables from the `main` branch. I updated the `.github/dependabot.yml` file to not specify a target branch, so that it will by default use the default branch (`main`).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,7 +2,6 @@ version: 2
 updates:
   - package-ecosystem: "pip"
     directory: "/"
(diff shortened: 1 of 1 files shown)


zschira (approved):
:+1:

Issue #506, opened 2025-12-11
Switch to geocod.io's Python client library


katie-lamb:
`geocodio`  has released a new official Python client library called [geocodio-library-python](https://github.com/Geocodio/geocodio-library-python). This could be a better replacement for [`pygeocodio`](https://github.com/bennylope/pygeocodio) which is a non-official Python client library.

Experiment with plugging in the official python library to see if it passes the ETL and CI.

This library requires updating to at least Python 3.11

Making this switch also seems to fix some of the mysterious geocoder changes in #507 . When this package switch happens, make sure that all of the spot fixes in that PR are addressed/removed.

- [x] upgrade to Python 3.12, use Ruff's `pyupgrade` to help
- [x] add `geocodio-library-python` as a dependency
- [x] check that DeSoto Parish is correctly geocoded in EIP infrastructure transform (see TODO in #507 )
- [x] check if Saratoga County is correctly geocoded in local opposition transform
- [x] fix unit test - random county and state gets geocoded to "Nome", may be able to use probability to help determine what is a good geocoded value
- [x] check if FYI locations table stays the same

PR #508, opened 2025-12-14
Python 3.13 update

katie-lamb:
This repo was previously on python 3.10. This PR upgrades it to Python 3.13. I started by using ruff's `pyupgrade` and then checked to make sure the ETL and CI passed. 

This also involved integrating ruff instead of black as the linter for this repo. A lot of the files in the diff are there because of `ruff` changes that needed to happen anyways, for example removing the `inplace` argument.

This PR is blocking us from switching to the Python geocodio library #506 , which will fix strange errors in the geocoder that have been spot fixed for now.

I know the diff is huge for this PR and it's difficult to walk through each of the changes, so I checked this by comparing the tables between the versions created with python 3.13 and python 3.10 to make sure they're the same. We should probably integrate (better) validation tests into this repo.

To Do:

- [x] compare data mart tables from `main` to this branch
- [x] debug differences between `counties_wide_format` table 
- [x] compare data warehouse tables from `main` to this branch
- [x] remove all use of the `inplace` argument

### DIFF
--- a/src/dbcp/transform/eip_infrastructure.py
+++ b/src/dbcp/transform/eip_infrastructure.py
@@ -135,12 +135,6 @@ def facilities_transform(raw_fac_df: pd.DataFrame) -> pd.DataFrame:
         "cancer_prevalence": "raw_percent_cancer_prevalence",
         "cancer_prevalence_percentile": "raw_percentile_cancer_prevalence",
         "county_fips_code_text": "raw_county_fips_code",
-        "location": "raw_location",
-        "facility_footprint": "raw_facility_footprint",
-        "epafrsid1": "raw_epa_frs_id_1",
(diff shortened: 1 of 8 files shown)


src/dbcp/transform/eip_infrastructure.py
```diff
@@ -139,12 +135,6 @@ def facilities_transform(raw_fac_df: pd.DataFrame) -> pd.DataFrame:
         "cancer_prevalence": "raw_percent_cancer_prevalence",
         "cancer_prevalence_percentile": "raw_percentile_cancer_prevalence",
         "county_fips_code_text": "raw_county_fips_code",
-        "location": "raw_location",
```
zschira: What are these being removed for?

PR #551, opened 2026-03-19
Fix deprecation warnings

katie-lamb:
We recently upgraded pandas and python, which led to a whole host of deprecation warnings popping up as we run the ETL. This PR addresses those warnings.

Closes #539


katie-lamb:
Hey @zschira I never quite finished / annotated this PR but this was just to get rid of all the deprecation warnings in the ETL after we updated the python/pandas/sqlalchemy versions. I think it's pretty much ready to go though so I'm handing it off.

PR #560, opened 2026-04-09
Add data warehouse docs

katie-lamb:
Add data warehouse docs and update docs to reflect current state of ETL.

In the future it would be great if we had more automated docs that pull from docstrings rather than needing to keep this data warehouse/mart page updated every time data changes, but I think this is important to get in quickly before doing a big rework.

- [x] add in detail from BQ core datasets google doc


katie-lamb:
@zschira I added a section to the docs with all the currently ETL'ed data warehouse tables. It's not the most detailed, but I think it's good to get something in quickly since we've let the docs languish for so long. I also tagged @e-belfer in case this is helpful for getting up to speed with the repo, but a lot of these tables aren't really in current development, so it's probably not worth much more than a skim.

docs/data-warehouse/_eia860m__changelog__generators.md
```diff
@@ -0,0 +1,27 @@
+# _eia860m__changelog__generators
+
+This table contains change-log style EIA 860M generator data sourced from Catalyst Cooperative's PUDL resources.
+This table is not intended for analysis and is largely just a building block for other EIA data warehouse tables (indicated by the leading underscore in the name.) Derived from [this PUDL table](https://catalystcoop-pudl.readthedocs.io/en/stable/data_dictionaries/pudl_db.html#core-eia860m-changelog-generators), this table is a changelog of monthly reported generator data. There is a record corresponding to the first instance of a generator and associated characteristics with a report_date column and a valid_until_date column. Whenever any of the reported EIA-860M data was changed for a record, there will be a new changelog record with a new report_date.
```
zschira: We should probably use the new `docs.catalyst.coop` link so this doesn't get outdated.

```suggestion
This table is not intended for analysis and is largely just a building block for other EIA data warehouse tables (indicated by the leading underscore in the name.) Derived from [this PUDL table](https://docs.catalyst.coop/pudl/en/latest/data_dictionaries/pudl_db.html#core-eia860m-changelog-generators), this table is a changelog of monthly reported generator data. There is a record corresponding to the first instance of a generator and associated characteristics with a report_date column and a valid_until_date column. Whenever any of the reported EIA-860M data was changed for a record, there will be a new changelog record with a new report_date.
```

docs/data-warehouse/airtable__manual_ordinances.md
```diff
@@ -0,0 +1,23 @@
+# airtable__manual_ordinances
+
+This table captures manually maintained county-level ordinance flags that supplement external ordinances datasets.
+
+## Table Details
+
+**Source:** Madrone Airtable manual ordinances base
```
zschira: What do you mean by "ordinances base"? I'm not sure if it's supposed to say "ordinances database" or if there's another meaning.

docs/data-warehouse/airtable__offshore_wind_cable_landing_association.md
```diff
@@ -0,0 +1,22 @@
+# airtable__offshore_wind_cable_landing_association
```
zschira, line 1: I should have brought this up in the renaming PR, and we don't need to change this in this PR, but it occurs to me that `association` describes a category of table and we should probably treat it like `changelog` in our naming, which would make this table name:

`airtable__association__offshore_wind_cable_landing`

docs/data-warehouse/eia860m__annual__generators.md
```diff
@@ -0,0 +1,24 @@
...
+
+**Grain:** One row per generator per annual report date
+
+**Primary key column(s):** `plant_id_eia`, `generator_id`, `report_date`
+
+**Purpose:** Latest annual generator attributes with county/state FIPS added.
+
+## Transformations
+
+The ETL filters the PUDL generator table ([out_eia__yearly_generators](https://catalystcoop-pudl.readthedocs.io/en/stable/data_dictionaries/pudl_db.html#out-eia-yearly-generators)) to the latest complete calendar year, converts date-like columns to datetimes, and adds state and county FIPS IDs from the generator state/county fields.
```
zschira:
```suggestion
The ETL filters the PUDL generator table ([out_eia__yearly_generators](https://docs.catalyst.coop/pudl/en/latest/data_dictionaries/pudl_db.html#out-eia-yearly-generators)) to the latest complete calendar year, converts date-like columns to datetimes, and adds state and county FIPS IDs from the generator state/county fields.
```

docs/data-warehouse/README.md
```diff
@@ -1,2 +1,95 @@
 # Data Warehouse
 
+This section documents the data warehouse tables.
```
zschira: I think we can remove this first line.

docs/README.md
```diff
@@ -3,3 +3,29 @@
...
+
+### Segment meanings
+
+| Segment | Meaning | Example values |
+| ---- | ---- | ---- |
+| `source_name` | The upstream source system or source family | `fyi`, `acp`, `airtable`, `census`, `eia860m` |
+| `private_flag` | Whether the table is intended for non-public use | `private` |
+| `time_interval` | The time cadence or snapshot style represented by the table | `annual`, `monthly`, `quarterly`, `yearly`, `changelog` |
+| `row_granularity_or_aggregation` | The unit of observation or aggregation level | `projects`, `generators`, `counties`, `iso_regions` |
+| `table_contents_description` | The specific subject, shape, or metric captured in the table | `long_format`, `wide_format`, `status_codes`, `resource_capacity` |
```
zschira: Are we still using the `long_format` / `wide_format` distinction in naming? I thought we had been moving away from that.

zschira (approved):
I've got a few questions and suggestions, but I'm good with you deciding how to handle them then merging when ready.

docs/data-warehouse/airtable__manual_ordinances.md
katie-lamb: Oh Airtable calls their databases "bases" which is kind of weird. I'll just take "base" out to reduce confusion potential

docs/data-warehouse/airtable__offshore_wind_cable_landing_association.md
katie-lamb: mmm yes good point, maybe i could move this change into the data mart PR

docs/README.md
katie-lamb: Yes, good point, removing this.

PR #561, opened 2026-04-14
FYI March Update

katie-lamb:
The following changes were made: 
* updated URI to new month of data
* added a fix for a project where the same resource was listed twice and the capacity of the project shouldn't be summed
* added new columns to the locations table for `gis_latitude`, `gis_longitude`, `gis_lat_long_specificity`
  * we don't do anything with lat, long values right now but there are a list of location table fixes in an issue that could potentially utilize these lat, long values.


zschira (approved):
Looks good!

PR #564, opened 2026-04-28
Switch dependency management to `uv`

zschira:
## Overview
Closes #537

To switch over to `uv` for dependency management, this pr does the following:

- Format `pyproject.toml` for `uv`
- Remove `requirements.txt` and move dependencies to `pyproject.toml`
- Update `Dockerfile` to install dev environment with `uv`

While making these changes, I also decided to update the `pre-commit` hooks. These are largely copied from `pudl` with some hooks removed that would not be relevant here.


pyproject.toml
```diff
@@ -1,42 +1,49 @@
...
+  "gridstatus~=0.29.1",
+  "joblib~=1.2.0",
+  "jupyterlab~=4.5.5",
+  "matplotlib~=3.10.8",
+  "numpy<3",
+  "pandas~=2.3.3",
+  "pandas-gbq~=0.34.0",
+  "pandera[io]~=0.26.1",
+  "plotly~=5.15.0",
+  "psycopg2~=2.9.11",
```
jdangerx:
```suggestion
  "psycopg2-binary~=2.9.11",
```

Avoids building from source which broke on my computer.

pyproject.toml
```diff
@@ -1,42 +1,49 @@
-[build-system]
-requires = ["setuptools>=61", "wheel"]
-build-backend = "setuptools.build_meta"
```
jdangerx, line 3: if we keep this or switch to `hatchling` we can keep the `branch-compare` entry-point too. not sure if you feel like that's necessary.

jdangerx (commented):
A few small tweaks:

* have *a* build system
* keep `branch-compare` runnable
* switch to `psycopg2-binary` so you don't need postgres installed locally to build from source 😅 

diff for your `git apply` needs:
```diff
diff --git a/pyproject.toml b/pyproject.toml
index bce20d1..9e6bb1a 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,7 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
 [project]
 name = "deployment-gap-model"
 version = "0.1.0"
@@ -27,7 +31,7 @@ dependencies = [
   "pandas-gbq~=0.34.0",
   "pandera[io]~=0.26.1",
   "plotly~=5.15.0",
-  "psycopg2~=2.9.11",
+  "psycopg2-binary~=2.9.11",
   "pyairtable>=2.3.3",
   "pydantic>=2.9,<2.13",
   "pydata-google-auth~=1.9.1",
@@ -41,9 +45,15 @@ dependencies = [
   "us~=3.2.0",
 ]
 
+[project.scripts]
+branch-compare = "scripts.branch_compare_helper:main"
+
 [project.urls]
 Homepage = "https://github.com/catalyst-cooperative/down_ballot_climate"
 
+[tool.hatch.build.targets.wheel]
+packages = ["src/dbcp", "src/scripts"]
+
 [tool.ruff]
 # Configurations that apply to both the `format` and `lint` subcommands.
 target-version = "py313"
```

jdangerx (changes requested):
Er that last one was supposed to be "request changes", not "comment." The blocking thing is the psycopg2 thing, it seems like hatchling would be a nice-to-have.

PR #565, opened 2026-05-11
Integrate FYI data for 2026-05-01

zschira:
## Overview
This PR integrates the newest month of FYI data. It also updates the FYI update notebook to use new tables.


src/dbcp/transform/fyi_queue.py
```diff
@@ -226,29 +244,18 @@ def _normalize_resource_capacity(fyi_df: pd.DataFrame) -> dict[str, pd.DataFrame
...
-            resource_capacity_df[
-                resource_capacity_df.duplicated(subset=["project_id", "resource"])
-            ]
-        )
-        <= n_expected_duped_resources
-    ), (
-        f"More than {n_expected_duped_resources} projects found with the same resource "
-        "listed twice in capacity_by_generation_type_breakdown. Ensure that their capacities should be summed. "
-        f"They have project IDs: {resource_capacity_df[resource_capacity_df.duplicated(subset=['project_id', 'resource'])].project_id}"
+    found_dupes = set(
```
zschira, line 249: I switched this to explicitly check for the duplicates that expect instead of looking for the number of duplicates. Also added a new duplicate that appears in the new data, `caiso-2113`. This is a case of two battery resources listed for the same project.

Also I made `EXPECTED_DUPLICATE_PROJECTS` a global variable within the module so I can mock it in tests.

notebooks/data_updates/interconnection_fyi/interconnection-fyi-update.ipynb
```diff
@@ -24,16 +24,7 @@
    "execution_count": 2,
```
zschira, line 24: A lot changed in this notebook. We've restructured all of the old tables since the last update, so I updated it to accommodate. It also used to validate FYI data using LBNL + gridstatus data, but we're no longer updating either of these data sources, so I removed those validations and tried to add some new sanity checks. We can probably think more about validations long term, but I just tried to get this working with the current data in a semi-reasonable way so we can get this update in.

e-belfer (approved):
All runs as expected and looks reasonable to me. Optional but other helpful verification steps to add to the notebook:

```
# Did any county lose more than one project?
combined_active_projects.groupby('county_id_fips')['_merge'].value_counts().reset_index().set_index('county_id_fips').value_counts()
```

```
# What were the changes in capacity by queue and status?
new_fyi.groupby(['power_market', 'queue_status'])['capacity_mw'].sum()-old_fyi.groupby(['power_market', 'queue_status'])['capacity_mw'].sum()
```
We see huge withdrawals in MISO and West. MISO seems on track with [reporting](https://www.rtoinsider.com/129776-miso-plans-change-accounting-practices-record-queue-exits/).

PR #571, opened 2026-06-17
June 2026 fyi update

zschira:
## Overview
This PR integrates the latest month of FYI data. It also includes a fix for the new uv build system that was causing problems with running jupyter notebooks.


src/dbcp/transform/fyi_queue.py
```diff
@@ -207,8 +207,6 @@ def parse_capacity(row):
     "caiso-2113",
     "ladwp-q57",
     "caiso-955",
-    "tucson-electric-power-94",
```
zschira, line 210: We have a set of expected duplicate projects and will raise an error if we don't find exactly those duplicates. These two disappeared this month. I manually checked the raw data, and it reflects this change.

Dockerfile
```diff
@@ -18,6 +18,7 @@ ENV PATH="/root/.local/bin/:$PATH"
 
 WORKDIR /app
 
+COPY src /app/src
```
zschira, line 21: We need to copy the `src` directory into the container to get the environment built correctly. In hindsight this seems obvious, but tricked me for a while because running the ETL / tests was working fine, but I couldn't `import dbcp` in a notebook.

notebooks/data_updates/interconnection_fyi/interconnection-fyi-update.ipynb
```diff
@@ -60,13 +60,15 @@
      "name": "stderr",
```
zschira, line 60: This is our notebook for sanity checking new data.

Dockerfile
krivard: nice catch!

krivard (approved):
🚀

PR #577, opened 2026-07-27
Revive counties wide

zschira:
# Overview
During our recent audit of tables, we had decided to remove a bunch of tables including the `counties_wide_format` table. However, we now know that Madrone is depending on this table for a dashboard, so this PR revives the old table (as well as a number of upstream tables).

Mostly this PR involves reviving old code and does basically nothing new. However, due to changes since this table was deprecated I've been indecisive about whether or not there are other changes we want to make before merging this PR. I think for now the best path would be to get this PR merged so we can update the table as needed and have a baseline to work off of, then get some clarification from Madrone how we want to handle the following questions.

# Open questions / Future work
## What tables / sources do we actually need?
`counties_wide_format` is a large table composited from many upstream sources. Many of these upstream sources were no longer being used and had to be brought back in to revive the counties wide table. If we can identify what they are actually relying on this table for, then we can narrow down which of the sources are actually necessary. We can also look at creating a smaller more focused table to suit their needs.

## Do any of the upstream sources need to be updated?
Once we have answered the above question, we should identify whether any of the sources are out of date due to lack of recent use and need to be updated.

## Can we change names / distribution locations
The new/old table does not conform to our current naming conventions and it would be nice to make that consistent, but it will require an update on Madrone's side. They are also pulling from `bigquery`, but we have been moving towards using `postgres` as our primary platform for distribution, so it would be nice to make this consistent as well.


src/dbcp/data_mart/fossil_infrastructure_projects.py
```diff
@@ -0,0 +1,31 @@
+"""Module to create a table of EIP fossil infrastructure projects for use in spreadsheet tools."""
```
e-belfer, line 1: Just a note that I looked into this some time back and we can certainly update this data, which is now being published programmatically on a regular schedule.

src/dbcp/etl.py
```diff
@@ -21,27 +24,44 @@
...
+    lbnl_transformed_dfs = dbcp.transform.lbnl_iso_queue.transform(lbnl_raw_dfs)
+
+    return lbnl_transformed_dfs
 
 
 def etl_fyi_queue() -> dict[str, pd.DataFrame]:
     """Interconnection.fyi ISO Queues ETL."""
-    file_paths = load_yml_file(DATA_DIR / "file_paths.yml")
-    fyi_raw_dfs = dbcp.extract.fyi_queue.extract(file_paths["fyi_queue"].item())
+    fyi_uri = "gs://dgm-archive/interconnection.fyi/interconnection_fyi_dataset_2026-07-01.csv"
```
e-belfer: You'll want to update this to conform to the new data input method.

src/dbcp/etl.py
```diff
@@ -150,15 +248,42 @@ def run_etl(funcs: dict[str, Callable], schema_name: SchemaName):
     logger.info(f"Successfully finished {schema_name.value} ETL.")
 
 
+def etl_offshore_wind() -> dict[str, pd.DataFrame]:
+    """ETL manually curated offshore wind data."""
+    # get the latest version of the offshore wind data from the candidate yaml file
+    projects_uri = "airtable/Offshore Wind Locations DBCP Version/Projects.json"
```
e-belfer, line 244: We should think about how to update this to get added into the timestamp table and conform to the new data input YAML format.

src/dbcp/etl.py
```diff
@@ -70,33 +90,111 @@ def etl_ncsl_state_permitting() -> dict[str, pd.DataFrame]:
...
 def etl_acp_projects() -> dict[str, pd.DataFrame]:
     """ETL ACP projects."""
     raw_dfs = dbcp.extract.acp_projects.extract()
     transformed = dbcp.transform.acp_projects.transform(raw_dfs)
     return transformed
 
 
+def etl_manual_ordinances() -> dict[str, pd.DataFrame]:
+    """ETL manually maintained ordinances."""
+    raw_dfs = dbcp.extract.manual_ordinances.extract()
```
e-belfer, line 183: Same as above.

e-belfer (changes requested):
My main request here is to update these changes to conform to the new data input YAML and show up in our new timestamp table as much as possible.

src/dbcp/data_mart/fossil_infrastructure_projects.py
zschira: Good to know, once we get details from Madrone on which fields they're actually using we can assess if keeping this updated is important.

src/dbcp/etl.py
zschira: Updated everything to use the new system

src/dbcp/etl.py
zschira: Fortunately, we archive all of the airtable stuff on GCS so I was able to just reuse the GCS date functionality.

src/dbcp/transform/file_modification.py
```diff
@@ -0,0 +1,132 @@
...
+    resp = requests.get(url, headers=headers, timeout=200)
+    commits = resp.json()
+    commit = commits[0]
+    # Prefer committer date since it reflects repository commit time
+    date_str = commit["commit"]["committer"]["date"]
+    return _parse_iso_z(date_str)
+
+
+def get_gcs_modified_time(
+    filepath: str, generation_num: str | None = None
```
zschira, line 48: The one change I had to make here is for the `airtable` based tables. We archive these in GCS and use object versioning so I added the `generation_num` to make sure we're getting the proper version.

e-belfer (approved):
Looks good and runs as expected!

PR #578, opened 2026-08-04
Update FYI data for August

e-belfer:
Add new Interconnection.FYI queue data up through August 2nd.


notebooks/data_updates/interconnection_fyi/interconnection-fyi-update.ipynb
cmgosnell, line 1: non-blocking: would it be simple to use the link from `file_paths.yml` instead?

cmgosnell (approved):
This all looks straightforward to me! Although I have not seen many (any?) of these updates so I am not well tuned for gremlins to look out for. The notebook updates seem mostly there for visibility? If yes it'd be nice to add comments about what to look out for in the future but that is super not blocking.

nonetheless this seems good.

PR #591, opened 2026-08-24
Refactor dev environment and builds/deployment

zschira:
## Overview
This PR simplifies our development environment, and removes a lot of cruft from our builds / deployments.

### Dev environment
I've removed everything related to docker in this PR, and we now just use `uv` directly on your machine. I've made this work by getting rid of `postgres` for local development, and instead writing to `duckdb`.

### Builds / deployment
I've split builds and deployment into 2 separate workflows (roughly inline with what we're doing in PUDL now). The `update-data` workflow runs the ETL and writes parquet files to versioned directories in GCS. The `publish-data` workflow takes versioned outputs and distributes data to postgres / BigQuery.

### Tooling
I've added some tooling in the `makefile` to make it easy to inspect local and remote data outputs. `duckdb` makes this very easy as it's happy to read from all of the data formats we are using.


.github/workflows/name: auto-release.yaml
zschira, line 1: This workflow generates a release that was used for the old deployment/build setup.

.github/workflows/update-data.yml
```diff
@@ -4,7 +4,7 @@ name: update-data
 on:
   push:
     branches:
-      - "main"
+      - "**"
```
zschira, line 7: By running this on every branch we can get rid of the `test-full-build` workflow. We don't actually upload outputs for non-`main` branches though

.github/workflows/update-data.yml
```diff
@@ -14,129 +14,33 @@ on:
...
-        description: "Release tag to deploy to prod (leave blank to use latest published release)"
-        required: false
-        default: ""
-  schedule:
-    - cron: 0 0 * * 0
-  release:
-    types: [published] # immediately deploy prod when a release is published
 
 jobs:
-  matrix_prep:
```
zschira, line 27: Tried to get rid of a lot of this complicated logic.

.github/workflows/update-data.yml
```diff
@@ -169,32 +61,24 @@ jobs:
...
-      # The google-github-actions/auth step is run as runner:docker,
-      # so we need to give the workspace back to runner:docker
-      - name: Give ownership of the workspace back to root
-        if: always()
-        run: sudo chown -R runner:docker $GITHUB_WORKSPACE
-
-      - name: Who owns the workspace?
-        if: always()
-        run: ls -ld $GITHUB_WORKSPACE
+      - name: Save version number associated with completed build
```
zschira, line 79: Versioned outputs get written to GCS in directories like `gs://dgm-outputs/{VERSION_UUID}/`. `VERSION_UUID` gets written to a text file and saved as an artifact.

src/dbcp/metadata/data_mart.py
```diff
@@ -233,141 +233,6 @@
     schema=schema,
 )
 
-iso_projects_wide_format = Table(
```
zschira, line 236: I also removed a bunch of metadata for deprecated tables.

src/dbcp/transform/file_modification.py
```diff
@@ -1,47 +1,42 @@
...
-        "User-Agent": "get-last-mod-time/1.0",
-    }
-
-    resp = requests.get(url, headers=headers, timeout=200)
-    commits = resp.json()
-    commit = commits[0]
-    # Prefer committer date since it reflects repository commit time
-    date_str = commit["commit"]["committer"]["date"]
-    return _parse_iso_z(date_str)
+    result = subprocess.run(  # noqa: S603
```
zschira, line 21: This is fixing a small bug where we were using the github api to get the date a file in the repo was modified, but this may give inaccurate results if you're working on a branch and modify the file on that branch.

src/dbcp/etl.py
```diff
@@ -185,11 +184,10 @@ def etl_manual_ordinances() -> dict[str, pd.DataFrame]:
     return transformed
 
 
-def write_to_postgres_and_parquet(
-    dfs: dict[str, pd.DataFrame], engine: sa.engine.Engine, schema_name: SchemaName
-):
-    """Write data mart tables from a schema to postgres and parquet."""
+def write_to_duckdb(dfs: dict[str, pd.DataFrame], schema_name: SchemaName):
```
zschira, line 187: During development, we just write to duckdb now.

.github/workflows/publish-data.yml
```diff
@@ -2,11 +2,57 @@
...
+  workflow_run:
+    workflows: ["update-data"]
+    types:
+      - completed
+    branches:
+      - main
   workflow_dispatch:
+    inputs:
+      run_id:
+        description: Github action run ID, which is the number at the end of the URL when looking at a run
```
jdangerx: This is "the run ID that generated the outputs we're about to publish"?

.github/workflows/publish-data.yml
```diff
@@ -2,11 +2,57 @@
...
+        required: true
+        type: string
 
 jobs:
   publish:
+    # Only run when manually triggered or fsspec archiver failed
```
jdangerx: This `if` reads to me as "Only run when manually triggered or when update-data workflow succeeded" which seems different from what this comment says, am I reading this right?

src/dbcp/transform/file_modification.py
jdangerx: blocking: The comment at the call-site implies that we can get rid of this whole function if we get rid of Docker, does that sound right to you?

```python
        # Query GitHub API for the most recent commit touching this path
        # We do this because the Docker build does not have the .git project
        # embedded within it, meaning that running git log is not an option.
        time = _github_latest_commit_date(repo_rel_path)

```

pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
   "click>=8.4.2",
   "coloredlogs~=15.0.1",
   "duckdb>=1.3.0",
+  "duckdb-sqlalchemy",
```
jdangerx: blocking: this guy needs a version, yeah?

README.md
```diff
@@ -40,16 +40,8 @@ uv sync --only-dev
...
-sibling ``mamba``. If you're not using ``conda`` environments, there are other
-ways to install the Google Cloud SDK explained in the link above.
-
-```
-conda install -c conda-forge google-cloud-sdk
-```
-
-Finally, use ``gcloud`` to establish application default credentials
+To authenticate the docker container with GCP install the [gcloud utilities](https://docs.cloud.google.com/sdk/docs/install-sdk#latest-version) on your
+computer. Once complete, use ``gcloud`` to establish application default credentials
```
jdangerx:
```suggestion
To authenticate with GCP install the [gcloud utilities](https://docs.cloud.google.com/sdk/docs/install-sdk#latest-version) on your
computer. Once complete, use ``gcloud auth application-default login`` to establish application default credentials
```

README.md
```diff
@@ -159,131 +103,73 @@ Now that we’ve built the image and set the environment variables run:
...
-```
+Note that we currently publish all data to a single schema in the postgres instances,
+so the `schema_name` can be dropped for these cases.
 
-starts a jupyter lab instance at `http://127.0.0.1:8888/`. If you have another jupyter service running at `8888` you can change the port by setting an environment variable in your shell before running this command:
 
-```
-export JUPYTER_PORT=8890
-```
+### make inspect_version
```
jdangerx, line 141: Love adding these dev experience tools! Out of scope suggestion: might be nice to do `make list-versions`, too. But not here/now.

jdangerx (changes requested):
Wayyy simpler than before, nice work! A couple cleanup guys but in general this looks great :)

README.md
zschira: Definitely! We could also make it display metadata about the version, which would be super helpful since the verion numbers are just `uuid`s so it can be hard to find a specific version sometimes.

.github/workflows/publish-data.yml
zschira: Yeah and I've expanded the comment now to make that more clear

.github/workflows/publish-data.yml
zschira: Oh yeah, I copied this from a workflow in the archiver repo, but didn't update the comment. Fixed now!

pyproject.toml
zschira: Fixed!

src/dbcp/transform/file_modification.py
zschira: We can get rid of using the github API and just use git directly, which is what I had already done here. I updated the comment now so it should be accurate.

jdangerx (approved):
🚢

PR #593, opened 2026-09-03
Update FYI with September data

e-belfer:
Monthly FYI update. Notable spike in PJM active capacity which tracks with the re-opening of the queue a few months ago, otherwise nothing particularly notable here.


src/dbcp/constants.py
```diff
@@ -214,7 +214,7 @@
         "type": "Fossil",
     },
     "Nuclear": {
-        "codes": [],
+        "codes": ["Fuel Cell + Gas + Nuclear", "Gas + Nuclear"],
```
e-belfer, line 217: These *could* also get mapped to gas instead, if that seems like more of the operative category here?

zschira: I think this is fine. We already split out by resource when we have enough info to do so, and we don't really look at nuclear / gas projects.

zschira (approved):
Looks good! I'm happy to merge this with my deployment branch and actually publish the updates.

PR #594, opened 2026-09-08
Track when tables are updated

### LINKED ISSUE #592: Track table-level last-updated time
## Overview

We want to show a little "last updated" line in the county dashboard for various data points.

## Success criteria

* [ ] each table we publish to postgres has a 'last-updated' column
* [ ] the last-updated column shows the last time the *table* was updated, *unless* row-level update times turn out to be easier.

zschira:
## Overview
This PR adds a new table called `madrone__deployment_metadata`. This table includes 1 row per table and columns indicating the build ID and timestamp when a table was last modified.

## Implementation
I noticed while implementing this that we were already logging the the build ID in `bigquery`, so I decided to use that here. Now while doing deployment to `bigquery`, we check these IDs, so we can compare each table to the previous version. After the comparison is complete, we will only update the table and metadata if the table has changed. The `load_tables_to_bigquery` function returns a `dataframe` with this metadata, so we can upload it to `postgres`.

### DIFF
--- a/src/dbcp/commands/publish.py
+++ b/src/dbcp/commands/publish.py
@@ -79,7 +79,7 @@ class DeploymentMetadata(pa.DataFrameModel):
 
     table_name: str = pa.Field(unique=True)
     last_modified_deployment_id: str
-    last_modified: pd.Timestamp
+    last_modified: pd.Timestamp = pa.Field(coerce=True)
 
 
 @pa.check_types
@@ -158,21 +158,6 @@ def load_tables_to_bigquery(
             # Construct the destination table
             table_ref = dataset_ref.table(table_name)
 
-            # delete table if it exists
-            client.delete_table(table_ref, not_found_ok=True)
-
-            # Load the Parquet file to BigQuery
-            job_config = bigquery.LoadJobConfig(
-                source_format=bigquery.SourceFormat.PARQUET,
-                write_disposition=bigquery.WriteDisposition.WRITE_TRUNCATE,
-            )
-            load_job = client.load_table_from_uri(
-                str(file), table_ref, job_config=job_config
-            )
-
-            logger.info(f"Loading {file.name} to {dataset_id}.{table_name}")
-            load_job.result()
-
             # Get the current deployed version of the table
             table = client.get_table(table_ref)
             current_version = table.labels.get("version")
@@ -184,8 +169,22 @@ def load_tables_to_bigquery(
             # Only update table if it's changed from previous deployment
             old_metadata = OutputMetadata.from_version(current_version)
             if not table_versions_equivalent(
-                old_metadata.output_directory / file.name, file
+                old_metadata.output_directory / schema.value / file.name, file
             ):
+                # Load the Parquet file to BigQuery
+                job_config = bigquery.LoadJobConfig(
+                    source_format=bigquery.SourceFormat.PARQUET,
+                    write_disposition=bigquery.WriteDisposition.WRITE_TRUNCATE,
+                )
+                load_job = client.load_table_from_uri(
+                    str(file), table_ref, job_config=job_config
+                )
+
+                logger.info(f"Loading {file.name} to {dataset_id}.{table_name}")
+                load_job.result()
+
+                # Reload table
+                table = client.get_table(table_ref)
                 labels = dict(table.labels or {})
                 labels["version"] = version
                 table.labels = labels
@@ -195,11 +194,15 @@ def load_tables_to_bigquery(
 
                 current_version = version
                 deployment_time = datetime.now()
+            else:
+                logger.info(
+                    f"{file.name} hasn't changed since previous deployment. Skipping upload."
+                )
 
             deployment_metadata.loc[len(deployment_metadata)] = {
                 "table_name": table_name,
                 "last_modified_deployment_id": current_version,
-                "last_modified": deployment_time,
+                "last_modified": pd.Timestamp(deployment_time),
             }
     return deployment_metadata
 
--- a/src/dbcp/commands/publish.py
+++ b/src/dbcp/commands/publish.py
@@ -21,9 +21,9 @@ from upath import UPath
 
 from dbcp.constants import DUCKDB_PATH
 from dbcp.helpers import (
+    check_table_versions_equivalent,
     get_postgres_engine,
     get_schema_sql_alchemy_metadata,
-    table_versions_equivalent,
     write_to_sql,
 )
 from dbcp.metadata import SchemaName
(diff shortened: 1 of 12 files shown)


src/dbcp/commands/publish.py
```diff
@@ -71,42 +74,58 @@ def upload_parquet_directory_to_gcs(
         )
 
 
+class DeploymentMetadata(pa.DataFrameModel):
```
zschira, line 77: Right now I've only defined the schema here in a `pandera` model, and it only gets published to `postgres`. That makes this table pretty inconsistent from how other tables are handled, but I think that's ok since this table only can be constructed while deploying data and is pretty categorically different from other tables.

That said, I'm open to suggestions if we want to try and make this more consistent.

src/dbcp/commands/publish.py
```diff
@@ -71,42 +74,58 @@ def upload_parquet_directory_to_gcs(
...
-                path=str(output_directory / schema.value / f"{table_name}.parquet")
-            ),
-            table_name=table_name,
-            engine=publish_engine,
-            schema_name=schema,
-            if_exists="replace",
-            remote=True,
-        )
-        logger.info(f"Successfully wrote table {table} to {target} postgres DB.")
+    for schema in SchemaName:
```
zschira, line 98: Rather than calling these deployment functions twice with each schema, I've just moved the iterating over schemas inside the functions to make it easier to create the deployment metadata table all at once.

src/dbcp/commands/publish.py
```diff
@@ -120,41 +139,72 @@ def load_tables_to_bigquery(
...
+            else:
+                logger.info(
+                    f"{file.name} hasn't changed since previous deployment. Skipping upload."
+                )
+
+            deployment_metadata.loc[len(deployment_metadata)] = {
+                "table_name": table_name,
+                "last_modified_deployment_id": current_version,
+                "last_modified": pd.Timestamp(deployment_time),
+            }
```
jdangerx, line 206: nit: this append feels awkward, might be nicer to have `deployment_metadata_records = []` up top, then have:

```
    deployment_metadata_records.append({...})
return DataFrame[DeploymentMetadata](deployment_metadata_records)
```

but certainly this is a nit

src/dbcp/commands/publish.py
```diff
@@ -71,42 +74,58 @@ def upload_parquet_directory_to_gcs(
...
-            schema_name=schema,
-            if_exists="replace",
-            remote=True,
-        )
-        logger.info(f"Successfully wrote table {table} to {target} postgres DB.")
+    for schema in SchemaName:
+        for table in get_schema_sql_alchemy_metadata(schema).sorted_tables:
+            table_name = table.name
+            # TODO: Figure out if county_wide + intermediate tables should be published to postgres
+            if "__" not in table_name:
```
jdangerx, line 102: nit: would be good to leave a note here about why `__` is the magic 'do publish this to postgres' marker

src/dbcp/commands/publish.py
jdangerx: I think that's OK but we should write down something like

* we only publish this deployment timing metadata to Postgres, because madrone only needs this for their postgres-backed dashboard
* if we were to want to extend this to BigQuery in the future, we'd probably add a `schema` column since BQ has tables under multiple schemas and Postgres only has one
* we don't go through SQLAlchemy for this table because it's only generated at publication time and only lives in one output database.

So that in the future this strange pattern isn't confusing to us

src/dbcp/commands/publish.py
```diff
@@ -71,42 +74,58 @@ def upload_parquet_directory_to_gcs(
...
+
+
+@pa.check_types
 def load_tables_to_postgres(
     output_directory: UPath,
-    schema: SchemaName,
     target: str,
+    deployment_metadata: DataFrame[DeploymentMetadata],
 ):
     """Load Parquet files from GCS to production postgres db.
```
jdangerx, line 91: Not always production, right, depends on target?

src/dbcp/commands/publish.py
```diff
@@ -71,42 +74,58 @@ def upload_parquet_directory_to_gcs(
...
+    )
 
 
+@pa.check_types
 def load_tables_to_bigquery(
-    output_directory: UPath, schema: SchemaName, target: str, version: str
-):
+    output_directory: UPath, target: str, version: str
+) -> DataFrame[DeploymentMetadata]:
     """Load Parquet files from GCS to BigQuery.
```
jdangerx, line 129: This could use a bit more detail about what we're doing, now: 

* diffing against old Parquet files & only updating BQ for files that are Different Enough
* keeping track of the version ID + timestamp of that version as a changelog, which later gets persisted to Postgres

Basically we're using the `version` label in BQ to help us generate a changelog, which we then push to Postgres. Feels a little fragile but if we document that well I'm OK with it.

src/dbcp/commands/publish.py
```diff
@@ -120,41 +139,72 @@ def load_tables_to_bigquery(
...
+                    write_disposition=bigquery.WriteDisposition.WRITE_TRUNCATE,
+                )
+                load_job = client.load_table_from_uri(
+                    str(file), table_ref, job_config=job_config
+                )
+
+                logger.info(f"Loading {file.name} to {dataset_id}.{table_name}")
+                load_job.result()
+
+                # Reload table
```
jdangerx, line 186:
```suggestion
                # Reload table metadata so we can update `version` label
```

src/dbcp/commands/publish.py
```diff
@@ -120,41 +139,72 @@ def load_tables_to_bigquery(
...
-    dataset_ref = client.dataset(dataset_id)
+    # Initialize deployment metadata
+    deployment_metadata = DeploymentMetadata.empty()
+
+    for schema in SchemaName:
+        # Get the BigQuery dataset
+        dataset_id = _get_published_schema_id(schema, target)
+        dataset_ref = client.dataset(dataset_id)
+
+        # Load each Parquet file to BigQuery
```
jdangerx, line 150:
```suggestion
```
We're not just "loading each Parquet file to BQ", we're "looking at each Parquet file, determining whether it needs to be uploaded, and then uploading" - I think we can get away with documenting that high level flow in the docstring.

src/dbcp/helpers.py
```diff
@@ -90,6 +91,30 @@ def get_bq_schema_from_metadata(
     return bq_schema
 
 
+def check_table_versions_equivalent(old_version: UPath, new_version: UPath) -> bool:
+    """Takes path to two parquet files of the same table and return a bool indicating if there are differences.
+
+    This function is meant to detect changes between deployments. It checks for major changes like
+    new, missing, or changed rows. For float columns it compares within a reasonable tolerance.
+    """
+    old_df = pd.read_parquet(str(old_version))
```
jdangerx, line 100: blocking: This breaks if we decide to publish a totally new table or rename a table, we should have some sort of "if no old table, return `True`" check.

jdangerx (changes requested):
I think this mostly works, which is good! Because of the weird BQ/Postgres schema mismatch and purgatory status of BQ in general, this is a bit of a house of cards. Which is fine, I think, but better documentation is a blocking concern otherwise we're going to be scratching our heads in a month's time.

There's also another small blocking change which is "how do we handle when the old Parquet file doesn't exist."

